PR #29, opened 2017-10-27
timer: fix timeout over about 524s do not work as expected

yangxingwu:


yangxingwu:
if we add a timer at level 2 when the cursor of level 1 is not 0, it will expire earlier than expected.

ywc689:
The timer carry problem is fixed in PR #470 .

ywc689:
> The timer carry problem is fixed in PR #470 .

Yes, I will reconsider it later.

PR #38, opened 2017-11-09
neigh: reply arp request only if tip is configured on the incoming interface

yangxingwu:
if we add a scope host route with dpip

```bash
./dpip route add 1.1.1.1/32 scope host dev dpdk0
```

then `arping` it

```bash
apping -I eth0 1.1.1.1
```

and we will get response

linux will normally ignore arp request if tip is not configured on the incoming interface, see `arp_ignore` for more details


beacer:
@yangxingwu any update about the review ?

yangxingwu:
I didn't see any comments and feedbacks...@beacer

ywc689:
We refactored ARP and Neighbor in ipv6 branch, and the problem is fixed. ipv6 branch will be merged into master branch soon. Thanks for your contribution!

PR #44, opened 2017-11-13
netif: packet capturing support by kni for debug.

kldeng:
Add flag to enable/disable packet capturing through kni device.
When enabled, all packets TX/RX will be copied and passed to KNI.
It's for debug only, and will affect the performance significantly.

### DIFF
--- a/src/netif.c
+++ b/src/netif.c
@@ -1789,15 +1789,36 @@ static inline void netif_tx_burst(lcoreid_t cid, portid_t pid, queueid_t qindex)
 {
     int ntx, ii;
     struct netif_queue_conf *txq;
+    unsigned i = 0;
+    struct rte_mbuf *mbuf_cloned = NULL;
+    struct netif_port *dev = NULL;
+
     assert(LCORE_ID_ANY != cid);
     txq = &lcore_conf[lcore2index[cid]].pqs[port2index[cid][pid]].txqs[qindex];
     if (0 == txq->len)
         return;
+
+    dev = netif_port_get(pid);
+    if (dev && (dev->flag & NETIF_PORT_FLAG_FORWARD2KNI)) {
+        for (; i<txq->len; i++) {
+            /*
+            * The rte_mbuf pending to transmit won't be modified any more.
+            * So, rte_pktmbuf_clone is ok.
+            */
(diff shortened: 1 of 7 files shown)


src/netif.c
```diff
@@ -2000,7 +2017,8 @@ static inline int netif_deliver_mbuf(struct rte_mbuf *mbuf,
     assert(dev != NULL);
 
     pt = pkt_type_get(eth_type, dev);
-    if (NULL == pt) {
+
+    if (!(dev->flag & NETIF_PORT_FLAG_FORWARD2KNI) && NULL == pt) {
```
beacer: if `err = pt->func(mbuf, dev);` returns `EDPVS_KNICONTINUE`, the mbuf will inject to `kni` again. pls also add the check `!(dev->flag & NETIF_PORT_FLAG_FORWARD2KNI)` to that place.
or the better way I think, to add a parameter `bool kni` to this API like
```
 static inline int netif_deliver_mbuf(struct rte_mbuf *mbuf,
                                      uint16_t eth_type,
                                      struct netif_port *dev,
                                      struct netif_queue_conf *qconf,
                                      bool kni)
```
Then if the mbuf is already copied to kni (in `lcore_job_recv_fwd`) , do not inject again.

src/netif.c
```diff
@@ -2069,6 +2088,20 @@ static void lcore_job_recv_fwd(void *arg)
                 eth_hdr = rte_pktmbuf_mtod(mbuf, struct ether_hdr *);
                 /* reuse mbuf.packet_type, it was RTE_PTYPE_XXX */
                 mbuf->packet_type = eth_type_parse(eth_hdr, dev);
+
+                /*
+                   * In NETIF_PORT_FLAG_FORWARD2KNI mode.
```
beacer: 1. comment indent
use
```
/*
 *
 */
```
instead of
```
/*
    *
    *
    */
```

2. be sure each line not exceed  80 chars.
 Yes, a lot of codes in `DPVS`, like netif.c donot follow this rule, but for new added codes, I think we should follow that.

kldeng:
hi, @beacer  
I realized that it's not elegant to close the previous PR but failed to find the way to reuse the previous PR before.  Now I got the correct way to reuse PR. :)

I modified the commit referring to your advice and also adding an indicator to show the current status of forward2kni flag.

PR #47, opened 2017-11-16
conf: fix invalid configuration 'netif_defs/device/tx/descriptor_number'

ywc689:
default value of netif_defs/device/tx/descriptor_number changed from 64->256
default value of netif_defs/device/tx/descriptor_number changed from 128->512

note that both rx and tx descriptor_number  are set to 1024 in /etc/dpvs.conf*.sample

### DIFF
--- a/conf/dpvs.conf.items
+++ b/conf/dpvs.conf.items
@@ -25,12 +25,12 @@ netif_defs {
         rx {
             #max_burst_size     32
             queue_number        6           <16, 0-16>
-            descriptor_number   128         <64, 16-8192>
+            descriptor_number   256         <256, 16-8192>
             rss                 ip          <tcp, tcp|ip>
         }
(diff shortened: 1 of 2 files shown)


conf/dpvs.conf.items
```diff
@@ -25,12 +25,12 @@ netif_defs {
...
             #max_burst_size     32
             queue_number        6           <16, 0-16>
-            descriptor_number   128         <64, 16-8192>
+            descriptor_number   128         <128, 16-8192>
             rss                 ip          <tcp, tcp|ip>
         }
         tx {
             queue_number        6           <16, 0-16>
-            descriptor_number   256         <128, 16-8192>
+            descriptor_number   256         <256, 16-8192>
```
beacer: 512 for some NIC ?

PR #98, opened 2018-01-16
Add a switch of ipv4_forward in ipv4_defs config section.

icymoon:
Set it to off, and dpvs won't forward packets that don't match the rules.

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -33,6 +33,9 @@
 #define INET_MAX_PROTS      256     /* cannot change */
 #define INET_DEF_TTL        64
 
+#define IPV4_FORWARD_DEF  false
+static bool ipv4_forward_switch = IPV4_FORWARD_DEF;
+
 static uint32_t inet_def_ttl = INET_DEF_TTL;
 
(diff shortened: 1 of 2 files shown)


src/ipv4.c
```diff
@@ -54,6 +57,15 @@ static void ipv4_default_ttl_handler(vector_t tokens)
     FREE_PTR(str);
 }
 
+static void ipv4_forward_handler(vector_t tokens)
+{
+    char *str = set_value(tokens);
+    assert(str);
+    if (!strncmp(str, "on", strlen("on")) || !strncmp(str, "ON", strlen("ON")))
```
beacer: You can use `strcasecmp`, and do not use 'n' version here, or the value "onABC" will also match.

src/ipv4.c
```diff
@@ -327,6 +340,11 @@ static int ipv4_forward(struct rte_mbuf *mbuf)
         goto drop;
     }
 
+    /* Drop packet if the switch is off */
+    if (! ipv4_forward_switch) {
```
beacer: remove space after `!`.

PR #102, opened 2018-01-19
Add a fullnat configure script sample.

icymoon:
Add a fullnat configure script sample.
Fix check_default_gw() function in snat's configure script sample.

### DIFF
--- a/scripts/config_fullnat_2arm_sample.sh
+++ b/scripts/config_fullnat_2arm_sample.sh
@@ -1,0 +1,178 @@
+#!/bin/sh
+#$Date: 2018-01-19
+#$ID: icymoon
+
+##### Basic configurations #####
+##### NOTE: modify these according to your environment #####
+WAN_IP_LIST=('192.168.10.4' '192.168.10.5') 
+WAN_TCP_PORT=80
+WAN_UDP_PORT=53
+WAN_NET='192.168.10.0/24'
+
+LAN_IP='172.16.10.2' #defined but not used
+LAN_PREF=24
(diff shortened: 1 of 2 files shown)


scripts/config_fullnat_2arm_sample.sh
```diff
@@ -0,0 +1,146 @@
+#!/bin/sh
+#$Date: 2018-01-19
+#$ID: icymoon
+
+##### Basic configurations #####
+##### NOTE: modify these according to your environment #####
+WAN_IP_LIST=('192.168.10.2' '192.168.10.4' '192.168.10.5') 
+WAN_PORT=80
+WAN_GW='192.168.10.0/24'
```
beacer: "Gateway" should not be `.0`, you can use `WAN_NET='192.168.10.0/24'`. And the same with `LAN_GW`.

scripts/config_fullnat_2arm_sample.sh
```diff
@@ -0,0 +1,146 @@
...
+    for d in ${WAN_IP_LIST[@]}; do
+        run_cmd "$DPIP_CMD addr show dev $WAN_DEV | grep inet \| grep $d/32 " \
+            "Check $d/32 on dev $WAN_DEV"
+  	done
+
+    run_cmd "$DPIP_CMD addr show dev $LAN_DEV | grep inet | grep $LAN_IP/$LAN_PREF " \
+        "Check $LAN_IP/$LAN_PREF on dev $LAN_DEV"
+}
+
+# Add default gw to $WAN_DEV
```
beacer: The comment is not consistent with codes.

PR #122, opened 2018-01-30
Don't load ip_vs kernel module and fix a segmentation fault.

icymoon:
1. Don't load kernel module in keepalived for dpvs.
2. Fix segmentation fault on starting keepalived without dpvs.
Thank you.

### DIFF
--- a/tools/keepalived/keepalived/check/check_data.c
+++ b/tools/keepalived/keepalived/check/check_data.c
@@ -42,6 +42,9 @@ alloc_ssl(void)
 void
 free_ssl(void)
 {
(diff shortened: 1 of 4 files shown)


tools/keepalived/keepalived/check/check_data.c
```diff
@@ -42,6 +42,9 @@ alloc_ssl(void)
 void
 free_ssl(void)
 {
+	if (!check_data)
```
beacer: indent, note keepalived and ipvsadm use `tab` for indent.

PR #135, opened 2018-02-26
Fix a fd leak in dpvs_running() 

icymoon:
Thanks.


src/ipvs/ip_vs_proto_icmp.c
```diff
@@ -132,6 +132,29 @@ static bool is_icmp_reply(uint8_t type)
...
+
+    ports[0] = icmp4_id(ich);
+    ports[1] = ich->type << 8 | ich->code;
+
+    /* ID may confict for diff host, should we use ID pool ? */
+    if (reverse) {
+        dp_vs_conn_fill_param(iph->af, iph->proto, &iph->daddr, &dest->addr, 
+                ports[1], ports[0], 0, param);
+    } else {
+        dp_vs_conn_fill_param(iph->af, iph->proto, &iph->daddr, &dest->addr,
```
beacer: It was `iph->saddr`, `iph->daddr` for `<CIP, VIP>`, here use `dest->addr` for VIP, it is not correct.
```
            dp_vs_conn_fill_param(iph->af, iph->proto,
                                  &iph->saddr, &iph->daddr,
                                  ports[0], ports[1], 0, &param);
```
Since it's not easy to eliminate "duplicated" codes, suggest not combine the codes with icmp_fill_param.

PR #145, opened 2018-03-11
vlan: turn spinlock to rwlock for better performance.

AveyondFly:


AveyondFly:
In fact there would be no performance improvement only with this commit, since the laddr_lock in dp_vs_laddrs_bind() completely destroy the multiple cores's advantages. However, let's make things better step by step.

beacer:
I ask the question just because some article shows, in some env,  `rwlock` may not have better performance than `spin_lock`, so I want some data to show, for the vlan module, if `rwlock` is really better or just theoretically better.

PR #146, opened 2018-03-12
ip_vs_core update:

dabinDong:
add "conn->dest->fwdmode != DPVS_FWD_MODE_SNAT" judge in function
    xmit_inbound_icmp and xmit_outbound_icmp for traceroute debug.

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -410,11 +410,15 @@ static int ipv4_rcv_fin(struct rte_mbuf *mbuf)
         }
         return ipv4_forward(mbuf);
     } else {
-        RTE_LOG(DEBUG, IPV4, "%s: input route has no dst\n", __func__);
-        route4_put(rt);
-        return EDPVS_KNICONTINUE; /* KNI may like it, don't drop */
+	RTE_LOG(DEBUG, IPV4, "%s: input route has no dst\n", __func__);
+        if (etype != ETH_PKT_HOST) {
+            /* multicast or broadcast */
+            route4_put(rt);
+            return EDPVS_KNICONTINUE; /* KNI may like it, don't drop */
+        }
+		/* send packet for bypass */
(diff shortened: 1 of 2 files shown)


src/ipv4.c
```diff
@@ -410,11 +410,15 @@ static int ipv4_rcv_fin(struct rte_mbuf *mbuf)
...
     } else {
-        RTE_LOG(DEBUG, IPV4, "%s: input route has no dst\n", __func__);
-        route4_put(rt);
-        return EDPVS_KNICONTINUE; /* KNI may like it, don't drop */
+	RTE_LOG(DEBUG, IPV4, "%s: input route has no dst\n", __func__);
+        if (etype != ETH_PKT_HOST) {
+            /* multicast or broadcast */
+            route4_put(rt);
+            return EDPVS_KNICONTINUE; /* KNI may like it, don't drop */
+        }
```
beacer: Ｉｆ pkt is "no route", means the dest is not for our own IP or KNI, and no idea how to "forward" it to other network. So why re-send the pkt "no route"?

src/ipvs/ip_vs_core.c
```diff
@@ -339,8 +339,9 @@ static int xmit_outbound_icmp(struct rte_mbuf *mbuf,
     struct ipv4_hdr *iph = ip4_hdr(mbuf);
 
     /* no translation needed for DR/TUN. */
-    if (conn->dest->fwdmode != DPVS_FWD_MODE_FNAT
-            && conn->dest->fwdmode != DPVS_FWD_MODE_NAT) {
+    if (conn->dest->fwdmode != DPVS_FWD_MODE_FNAT && 
+        conn->dest->fwdmode != DPVS_FWD_MODE_NAT  &&
```
beacer: it's correct to add NAT here.

PR #148, opened 2018-03-16
free mbuf if packet processing failed

yangxingwu:


beacer:
Next time, pls use the format "module: description", like
`netif: fix mbuf leak when pkt not valid to pass kni`

PR #175, opened 2018-04-27
remove some redundant pieces of code.

icymoon:
Thanks.

### DIFF
--- a/src/netif.c
+++ b/src/netif.c
@@ -1940,7 +1940,7 @@ int netif_hard_xmit(struct rte_mbuf *mbuf, struct netif_port *dev)
     lcoreid_t cid;
     int pid, qindex;
     struct netif_queue_conf *txq;
-    struct netif_ops *ops = dev->netif_ops;
+    struct netif_ops *ops;
     int ret = EDPVS_OK;
 
     if (unlikely(NULL == mbuf || NULL == dev)) {
@@ -1949,6 +1949,7 @@ int netif_hard_xmit(struct rte_mbuf *mbuf, struct netif_port *dev)
         return EDPVS_INVAL;
     }
 
+    ops = dev->netif_ops;
     if (ops && ops->op_xmit)
         return ops->op_xmit(mbuf, dev);
 
--- a/src/sa_pool.c
+++ b/src/sa_pool.c
@@ -506,14 +506,14 @@ int sa_fetch(struct netif_port *dev, const struct sockaddr_in *daddr,
     int err;
     assert(saddr);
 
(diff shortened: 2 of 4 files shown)


src/sa_pool.c
```diff
@@ -506,31 +506,29 @@ int sa_fetch(struct netif_port *dev, const struct sockaddr_in *daddr,
     int err;
     assert(saddr);
 
-    if (saddr->sin_addr.s_addr != INADDR_ANY && saddr->sin_port != 0)
+    if (saddr &&  saddr->sin_addr.s_addr && saddr->sin_addr.s_addr != INADDR_ANY && saddr->sin_port != 0)
```
beacer: 1. `INADDR_ANY` is zero, so  `saddr->sin_addr.s_addr` is not needed.
2. actually `if (saddr->sin_addr.s_addr != INADDR_ANY && saddr->sin_port != 0)` is OK, because here's an `assert(saddr)` has already make sure `saddr` is not NULL.

src/sa_pool.c
```diff
@@ -506,31 +506,29 @@ int sa_fetch(struct netif_port *dev, const struct sockaddr_in *daddr,
...
-        if (!ifa)
-            return EDPVS_NOTEXIST;
-
-        if (!ifa->this_sa_pool) {
-            RTE_LOG(WARNING, SAPOOL, "%s: fetch addr on IP without pool.",
-                    __func__);
-            inet_addr_ifa_put(ifa);
-            return EDPVS_INVAL;
-        }
+    ifa = inet_addr_ifa_get(AF_INET, dev,
```
beacer: `if (saddr->sin_addr.s_addr)` should be reserved.
This section should be kept not  changed I think, it's for the case only source IP is assigned but not port.

src/tc/sch_tbf.c
```diff
@@ -267,12 +267,14 @@ static void tbf_reset(struct Qsch *sch)
...
-    struct tbf_sch_priv *priv = qsch_priv(sch);
+    struct tbf_sch_priv *priv;
     struct tc_tbf_qopt *qopt = arg;
 
-    if (!sch && sch->ops != &tbf_sch_ops)
+    if (!sch || sch->ops != &tbf_sch_ops)
         return EDPVS_INVAL;
 
+    priv = qsch_priv(sch);
+
```
beacer: ok for this function.

src/netif.c
```diff
@@ -1949,6 +1949,7 @@ int netif_hard_xmit(struct rte_mbuf *mbuf, struct netif_port *dev)
         return EDPVS_INVAL;
     }
 
+    ops = dev->netif_ops;
```
beacer: ok for this function.

PR #177, opened 2018-05-07
version: make it possibe to get version info during runtime

kldeng:
make it possibe to get version info during runtime.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -18,6 +18,7 @@
 #define _GNU_SOURCE
 #include <pthread.h>
 #include <assert.h>
+#include <getopt.h>
 #include "pidfile.h"
 #include "dpdk.h"
 #include "common.h"
@@ -73,6 +74,63 @@ static int set_all_thread_affinity(void)
     return 0;
 }
 
+static void dpvs_usage(const char *prgname)
+{
+    printf("\nUsage: %s ", prgname);
+    printf("DPVS application options:\n"
+            "   -v  version     display DPVS version info\n"
+            "   -h  help        display DPVS help info\n"
+    );
+}
+
+static int parse_app_args(int argc, char **argv)
+{
+    const char *short_options = "vh";
+    char *prgname = argv[0];
+    int c, ret = -1;
+
+    const int old_optind = optind;
+    const int old_optopt = optopt;
+    char * const old_optarg = optarg;
+
+    struct option long_options[] = {
+        {"version", 0, NULL, 'v'},
+        {"help", 0, NULL, 'h'},
+        {NULL, 0, 0, 0}
+    };
+
+    optind = 1;
+
+    while ((c = getopt_long(argc, argv, short_options, long_options, NULL)) != -1) {
+        switch (c) {
+            case 'v':
+                fprintf(stderr, "dpvs version: %s, build on %s\n",
+                        DPVS_VERSION,
+                        DPVS_BUILD_DATE);
+                exit(EXIT_SUCCESS);
+            case 'h':
+                dpvs_usage(prgname);
+                exit(EXIT_SUCCESS);
+            case '?':
+            default:
+                dpvs_usage(prgname);
+                exit(EXIT_FAILURE);
+        }
+    }
+
+    if (optind > 0)
+        argv[optind-1] = prgname;
+
+    ret = optind - 1;
+
+    /* restore getopt lib */
+    optind = old_optind;
+    optopt = old_optopt;
+    optarg = old_optarg;
+
+    return ret;
+}
+
 int main(int argc, char *argv[])
 {
     int err, nports;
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -86,7 +86,10 @@ int main(int argc, char *argv[])
 
     /* check if dpvs is running and remove zombie pidfile */
     if (dpvs_running(DPVS_PIDFILE)) {
-        fprintf(stderr, "dpvs is already running\n");
+        fprintf(stderr,
```
beacer: If the binary is changed for example upgraded, the version shown by this line is not the `DPVS` version which is actually running. So user may get confused.

kldeng: Not really get it. Shouldn't the binary be re-compiled and restarted when upgrades. So the user will get the version actually running. Please correct me if I am wrong. Thanks.

beacer: Assume the old version is 1.0, and it's already running. 
New image is 1.1, now try to execute 1.1, then it  will show
```
dpvs is already running
version 1.1, build on ....
```
User may think dpvs 1.1 is already running, but actually it's 1.0.

kldeng: yes, you're right. I've modified the PR by adding a application arguments parse before the EAL ones. Please review it, thanks!

src/main.c
```diff
@@ -73,6 +74,62 @@ static int set_all_thread_affinity(void)
...
+
+    while((c = getopt_long(argc, argv, short_options, long_options, NULL)) != -1) {
+        switch(c) {
+            case 'v':
+                fprintf(stderr, "dpvs version: %s, build on %s\n", DPVS_VERSION, DPVS_BUILD_DATE);
+                exit(EXIT_SUCCESS);
+            case 'h':
+                dpvs_usage(prgname);
+                exit(EXIT_SUCCESS);
+            case '?':
```
beacer: getopt_long return the case `?` equals to `default`, or 'default' branch will never get hit.
```
case '?':
default:
    dpvs_usage(prgname);
    exit(...);
```

src/main.c
```diff
@@ -73,6 +74,62 @@ static int set_all_thread_affinity(void)
...
+
+    struct option long_options[] = {
+        {"version", 0, NULL, 'v'},
+        {"help", 0, NULL, 'h'},
+        {NULL, 0, 0, 0}
+    };
+
+    optind = 1;
+
+    while((c = getopt_long(argc, argv, short_options, long_options, NULL)) != -1) {
```
beacer: Need space between `while` and `(`
```
while ((c = ...) ..) {
}
```

src/main.c
```diff
@@ -73,6 +74,62 @@ static int set_all_thread_affinity(void)
...
+        switch(c) {
```
beacer: spaces,
```
switch (c) {
```

src/main.c
```diff
@@ -73,6 +74,62 @@ static int set_all_thread_affinity(void)
...
+
+static int parse_app_args(int argc, char **argv)
+{
+    const char *short_options = "vh";
+    char *prgname = argv[0];
+    int c, ret = -1;
+
+    const int old_optind = optind;
+    const int old_optopt = optopt;
+    char * const old_optarg = optarg;
```
beacer: `const char *old_optarg = ...`

src/main.c
```diff
@@ -73,6 +74,62 @@ static int set_all_thread_affinity(void)
...
+        }
+    }
+
+    if (optind > 0)
+        argv[optind-1] = prgname;
+
+    ret = optind - 1;
+
+    /* restore getopt lib */
+    optind = old_optind;
```
beacer: If some argument are consumed, is old index still correct ?

src/main.c
```diff
@@ -84,9 +141,26 @@ int main(int argc, char *argv[])
...
+        exit(EXIT_FAILURE);
+    }
+    argc -= err, argv += err;
+
     /* check if dpvs is running and remove zombie pidfile */
     if (dpvs_running(DPVS_PIDFILE)) {
-        fprintf(stderr, "dpvs is already running\n");
+        fprintf(stderr, 
+            "dpvs is already running\ndpvs version: %s, build on %s\n", 
+            DPVS_VERSION,
```
beacer: The question I've raised still not get resolved, if print version here, user may get wrong version for old dpvs which is already running. Should not print version here I think.

src/main.c
kldeng: Done.

src/main.c
kldeng: char *const is a const pointer to char while const char * is a pointer to const char. We need the pointer not the value to be unchanged. We can also see it at eal_parse_args.

src/main.c
kldeng: Yes. old_optind won't change after that. And all eal arguments parse function will reset optind to 1 as double assurance.

PR #184, opened 2018-05-20
add dpdk 17.11 support

xiangp126:


src/netif.c
```diff
@@ -3777,6 +3777,23 @@ int netif_lcore_start(void)
...
+    const char *p_end = str + len - 1;
+    const char *p_begin = p_end;
+
+    /* bond device did not end up with number */
+    if (! (*p_end >= '0' && *p_end <= '9')) {
+        return -1;
+    }
+    while ((*p_begin >= '0' && *p_begin <= '9') && p_begin >= str) {
+        --p_begin;
+    }
```
beacer: 1. use `sscanf` to make the code clear and easy to handle other bad inputs.
2. generate the "net_bondingXX" in this function, so that if DPDK change the format some day, only this func need be modified.

```
static int obtain_dpdk_bond_name(const char *orig, char *name, size_t size)
{
        char str[DEVICE_NAME_MAX_LEN];
        unsigned num;

        if (!orig || sscanf(orig, "%[a-zA-Z]%u", str, &num) != 2)
                return EDPVS_INVAL;

        /*
         * DPDK need bonding device name start with "net_bonding"
         * to match the driver.
         */
        snprintf(name, size, "net_bonding%u\n", num);

        return EDPVS_OK;
}
```

xiangp126:
diff been updated

PR #192, opened 2018-06-03
Fixed compile errors on gcc7.

haosdent:
Got following errors when compile dpvs with gcc7.

```
src/kni.c:68:53: error: ‘.kni’ directive output may be truncated writing 4 bytes into a region of size between 1 and 32 [-Werror=format-truncation=]
         snprintf(conf->name, sizeof(conf->name), "%s.kni", dev->name);
```

```
gcc -o ../bin/keepalived `find core check vrrp libipvs-2.6 ../lib -name '*.[oa]'` ../../../src/common.o -lssl -lcrypto -lcrypt   -ldl -lnuma
/usr/bin/x86_64-linux-gnu-ld: cannot open output file ../bin/keepalived: No such file or directory
```

In this patch, we disable `format-truncation` warning via refer to
http://dpdk.org/dev/patchwork/patch/25037/ .


src/config.mk
```diff
@@ -33,3 +33,7 @@ CFLAGS += -D DPVS_MAX_LCORE=64
 #CFLAGS += -D DPVS_CFG_PARSER_DEBUG
 #CFLAGS += -D NETIF_BONDING_DEBUG
 #CFLAGS += -D CONFIG_TC_DEBUG
+
+GCC_MAJOR = $(shell echo __GNUC__ | $(CC) -E -x c - | tail -n 1)
+GCC_MINOR = $(shell echo __GNUC_MINOR__ | $(CC) -E -x c - | tail -n 1)
+GCC_VERSION = $(GCC_MAJOR)$(GCC_MINOR)
```
beacer: "no newline at end of file".

PR #194, opened 2018-06-05
Updated docs to include default route setting.

haosdent:


beacer:
Default route is not mandatory in the test environment for README and tutorial.
Furthermore, when fast-xmit is enabled, default may not needed, but using default route is better for some env.

haosdent:
Noted, thx @beacer Do you have plans to support fast-xmit

beacer:
fast-xmit is already supported, pls check `ip_vs_xmit.c`.

PR #209, opened 2018-06-20
neigh:fix bug: inconsistent neighbour static flag

mscbg:


ba0f3:
smth is still wrong, `arp_timeout` still not being used

PR #215, opened 2018-07-04
version: get version from git tags or commit id if no tag exists.

kldeng:
Get DPVS version from git tags or commit id if no tag exists instead of maintaining the VERSION shell script.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -23,7 +23,12 @@
 
 TARGET := dpvs
 
-VERSION_STRING := $(shell ./VERSION)
+ifneq ("$(wildcard VERSION)","")
+    VERSION_STRING := $(shell ./VERSION)
+else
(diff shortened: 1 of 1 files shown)


src/Makefile
```diff
@@ -23,7 +23,7 @@
 
 TARGET := dpvs
 
-VERSION_STRING := $(shell ./VERSION)
+VERSION_STRING := $(shell git describe --tags --always)
```
beacer: It's good idea that using command get version automatically, however some scripts may need VERSION file to parse the DPVS version on deployed machine, which has no source code installed.
We can change that scripts to use 'dpvs --version', but for now, we need compatible with these scripts.

so 
1. pls check VERSION and use file if exist, if not exist use git command.
2. reserve VERSION file for compatibility.

kldeng: Done and re-pulled the rqeuest.

PR #235, opened 2018-07-28
ipvs: __dp_vs_update_dest protected in dp_vs_edit_dest.

mfhw:
New connections are scheduling dest based on wrr/wlc will be affected
while we are updating the dest weight if not protected. dp_vs_add_dest
don't need to care it, since the dest updated by __dp_vs_update_dest is
still not in the schedule list.


ywc689:
Per-lcore Service/Dest/Laddr is adopted in pr #448. Locking is no need anymore.

PR #259, opened 2018-08-31
Under the snat+gre senceario, icmp will have problems of inconsistenc…

weiyanhua100:
…y of receiving and sending ports.

  The specific solution：
  1.All ICMP messages received from tunnel are sent to the 1 core for processing.
  2.The RSS of the external network port is configured as TCP/UDP

### DIFF
--- a/conf/dpvs.conf.sample
+++ b/conf/dpvs.conf.sample
@@ -227,6 +227,11 @@ ipv4_defs {
     }
 }
 
+! dpvs tunnel config
+tunnel_defs {
+    forward_icmp_use_first_core
+}
+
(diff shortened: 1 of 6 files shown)


ywc689:
Does the patch work?  I checked the problem, and think it is caused by RSS in essence. For ICMP, FDIR does not work,  and RSS is responsible for Rx queue assignment policy if RSS hash function is set to IP. Thus inbound packets may go to the different worker from the outbound packets by whom the DPVS connection is initialized. Sending outbound packets from tunnel to the first worker does not fix the session miss caused by inbound packets.

PR #266, opened 2018-09-04
neighbour promote

mscbg:
1,neighbour support per core
2,neighbour show support per core

### DIFF
--- a/src/neigh.c
+++ b/src/neigh.c
@@ -631,14 +631,9 @@ int neigh_output(int af, union inet_addr *nexhop,
             neigh_fill_mac(neighbour, m, NULL, port);
             netif_xmit(m, neighbour->port);
 
-            /* 
-             * state should change before send mbuf
-             * PROBE state will not send multicast, just send unicast
-             * then state_confirm can be a loop
-             */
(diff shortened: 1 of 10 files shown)


src/neigh.c
```diff
@@ -888,32 +898,83 @@ static int neigh_sockopt_get(sockoptid_t opt, const void *conf,
...
-                    neigh_fill_param(&array->addrs[off++], entry);
-                }
-            }
+    /* copy neigh_array*/
+    array->neigh_nums = neigh_nums_g;
+    list_for_each_entry(cur, &reply->mq, mq_node) {
+        array_msg = (struct dp_vs_neigh_conf_array *)(cur->data);
+        memcpy(&array->addrs[off], &array_msg->addrs, 
+               array_msg->neigh_nums * sizeof(struct dp_vs_neigh_conf));
+        off += array_msg->neigh_nums;
```
ywc689: All slaves should have the same neighbor entries ? Is it necessary to output so much duplicated information?

mscbg: 每个核上的邻居状态可能不一样，也有可能有某个核因为某bug导致无法获取邻居。方便排障定位问题。

PR #277, opened 2018-09-13
ipvs: add nat-mode support in single-core dpvs

andrewhit:
o In full-nat mode, the link address is added via local_address_group{}
  and laddr_group_name configured under virtual_server {}.
o However within non-full-nat mode, like NAT-mode, we fail to configure
  the link address via static_address{} because dpdk_port is set to NULL
  within start_vrrp().


andrewhit:
Here is the test result.

```
static_ipaddress {
   10.10.127.66/26 dev eth3
}

local_address_group laddr_g1 {
    10.10.127.68 eth3  # use DPDK interface
}

root@n10-023-013:/var/log# dpip addr show
inet 10.10.127.68/32 scope global eth3
     valid_lft forever preferred_lft forever sa_used 0 sa_free 4294950896 sa_miss 0
inet 10.10.127.66/26 scope global eth3
     valid_lft forever preferred_lft forever
```

PR #279, opened 2018-09-18
match zero port services to enable dmz

kaseiwang:
it seems DMZ mode are broken in current version, because of  __dp_vs_service_get doesn't match port zero in DMZ forward rules.

### DIFF
--- a/src/ipvs/ip_vs_service.c
+++ b/src/ipvs/ip_vs_service.c
@@ -31,6 +31,7 @@
 #include "neigh.h"
 
 static int dp_vs_num_services = 0;
+static rte_atomic32_t dp_vs_num_nullservices;
 
 /**
  * hash table for svc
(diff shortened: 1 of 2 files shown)


src/ipvs/ip_vs_service.c
```diff
@@ -113,10 +113,11 @@ struct dp_vs_service *__dp_vs_service_get(int af, uint16_t protocol,
...
         if ((svc->af == af)
             && inet_addr_equal(af, &svc->addr, vaddr)
-            && (svc->port == vport)
             && (svc->proto == protocol)) {
-                rte_atomic32_inc(&svc->usecnt);
-                return svc;
+                if ((svc->port == vport) || (svc->port == 0)) {
+                    rte_atomic32_inc(&svc->usecnt);
+                    return svc;
+                }
```
mscbg: function dp_vs_set_svc use __dp_vs_service_get. If dpvs has service 192.168.1.1:0 and we use command ipvasdm -ln -t 192.168.1.1:80, ipvsadm will show service wrong

kaseiwang:
yeah… Maybe we should add a boolean parameter for marching zero port or
not. Will rewrite this patch.

PR #289, opened 2018-09-27
ipvs: add nat-mode support in multi-cores node.

andrewhit:
o During the system boots up, the below resource is pre-allcoated.
  - per-socket based global connetion redirect hash table;
  - per-socket based connection redirect cache;
  - each lcore allocates the respective packet forward rings for each other

o When a connection is created and hashed in nat-mode, the related redirect
  is allocated and hashed accordingly.

o When a connection expires to be unhashed and freed in nat-mode, the related
  redirect is unhashed and freed accordingly.

  then check if it matches any connection redirect entry. If match, enqueue the
  packet into the packet forward ring of the rediret owner core; otherwise,

  the packet forward rings owned by the current lcore and process them
  accordingly.

### DIFF
--- a/tools/keepalived/keepalived/check/ipwrapper.c
+++ b/tools/keepalived/keepalived/check/ipwrapper.c
@@ -197,16 +197,20 @@ init_service_vs(virtual_server_t * vs)
 			SET_ALIVE(vs);
 	}
 
-	/*Set local ip address in "FNAT" mode of IPVS */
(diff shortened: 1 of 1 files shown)


ywc689:
The redirect table seems a pretty good method to solve NAT-forwarding single worker defect. But the redirect table is a global hash table, and  must be locked before access/modified. I wonder how much negative influence it has on the overall performance behavior of DPVS.

andrewhit:
For the convenience of code reviewing, the above pull request is replaced by https://github.com/iqiyi/dpvs/pull/328.

PR #303, opened 2018-10-11
ipvs: match zero port for dmz mode

kaseiwang:


ywc689:
Tools such as `iptables` is supported by DPVS now. We have no way to mark dpvs packets, so fw-mark service is not supported. We may consider the feature later if needed.

kaseiwang:
This feature doesn't need fw-mark. With this patch, we could add tcp services without specific port. It's more like assigning an EIP to realserver. For example, add such a service:
```
./ipvsadm -A -t ${EIP}:0 -s rr -p
./ipvsadm -a -t ${EIP}:0 -r ${RS} -b
./ipvsadm --add-laddr -z ${LIP} -t ${EIP}:0 -F dpdk0
```

PR #304, opened 2018-10-16
convert tunnel id to dev port id in tunnel mode, because the virtual …

weiyanhua100:
…service iif field is the physical port.

### DIFF
--- a/src/ipvs/ip_vs_service.c
+++ b/src/ipvs/ip_vs_service.c
@@ -29,6 +29,8 @@
 #include "netif.h"
 #include "assert.h"
 #include "neigh.h"
+#include "ip_tunnel.h"
+#include "vlan.h"
 
 static int dp_vs_num_services = 0;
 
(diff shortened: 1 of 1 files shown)


src/ipvs/ip_vs_service.c
```diff
@@ -210,6 +226,16 @@ __dp_vs_svc_match_get(int af, const struct rte_mbuf *mbuf)
...
+
+            out_dev = netif_port_get(oif);
+            if (!out_dev)
+                return NULL;
+
+            if (out_dev->type == PORT_TYPE_TUNNEL) {
+                tnl = netif_priv(out_dev);
+                if (tnl && tnl->link)
+                    oif = tnl->link->id;
+            }
```
ywc689: `oif` is from route lookup. I think changes for it is not needed.

src/ipvs/ip_vs_service.c
```diff
@@ -218,7 +244,7 @@ __dp_vs_svc_match_get(int af, const struct rte_mbuf *mbuf)
         if (svc->af == af && svc->proto == iph->next_proto_id &&
             __svc_in_range(af, &saddr, ports[0], &m->srange) &&
             __svc_in_range(af, &daddr, ports[1], &m->drange) &&
-            (!idev || idev->id == mbuf->port) &&
+            (!idev || idev->id == iif) &&
```
ywc689: Tunnel device is just a case for the problem only. Other virtual devices, such as vlan device, have different device id from the  id of `mbuf->port`.  This is not a completed fix.

src/ipvs/ip_vs_service.c
weiyanhua100: 现在还没有这个应用场景，将其去掉

src/ipvs/ip_vs_service.c
weiyanhua100: 添加了vlan虚拟端口的处理。

ywc689: Sorry, i was not aware `mbuf->id` is translated to the virtual device's port id in its receive function(i.e.  `p_tunnel_rcv`, `vlan_rcv`) just now. I think it can be matched anyway if `m->iifname` is set to tunnel/vlan device's name.  Use physical device name in `m->iifname` instead of virtual device name is not easy to understand, and may arise safety problem, even if it simplifies the config work when a lot many virtual devices existing. I do not think it needs to change here.

ywc689:
Use src-range instead in the case of many virtual devices existing. Use the underlying device for match is not a graceful way.

PR #317, opened 2018-11-05
do not update checker's state if failed to perform server's command

ytwang0320:
Health checker's state will always update even if the commands to DPVS fail to perform for some reasons.
It will lead the states of two ends out of sync. For example, the weight of RS is always 0 even though the health checker is alive.


tools/keepalived/keepalived/check/ipwrapper.c
```diff
@@ -496,40 +503,62 @@ svr_checker_up(checker_id_t cid, real_server_t *rs)
...
-			if (*id == cid) {
-				free_list_element(l, e);
-				/* If we don't break, the next iteration will trigger
-				 * a SIGSEGV.
-				 */
-				break;
-			}
+		remove_failed_checker_list(cid, rs);
+		if (LIST_SIZE(rs->failed_checkers) == 0) {
+			if(!perform_svr_state(alive, vs, rs))
```
ywc689: `if(` ->`if (`

tools/keepalived/keepalived/check/ipwrapper.c
```diff
@@ -496,40 +503,62 @@ svr_checker_up(checker_id_t cid, real_server_t *rs)
...
 	/* Handle not alive state */
 	else {
-		id = (checker_id_t *) MALLOC(sizeof(checker_id_t));
-		*id = cid;
-		list_add(l, id);
-		if (LIST_SIZE(l) == 1)
-			perform_svr_state(alive, vs, rs);
+		add_failed_checker_list(cid, rs);
+		if (LIST_SIZE(rs->failed_checkers) == 1) {
+			if(!perform_svr_state(alive, vs, rs))
```
ywc689: `if(` -> `if (`

PR #318, opened 2018-11-05
fix possible crash when exec ipvsadm -ln

yangxingwu:


src/ipvs/ip_vs_service.c
```diff
@@ -921,6 +921,8 @@ static int dp_vs_get_svc(sockoptid_t opt, const void *user, size_t len, void **o
         case DPVS_SO_GET_VERSION:
             {
                 char *buf = rte_zmalloc("info",64,0);
+                if (NULL == buf)
```
ywc689: `if (NULL == buf)` -> `if (unlikely(NULL == buf))`
All branches below need similar change.

yangxingwu: done!

thanks ywc :)

PR #319, opened 2018-11-05
fix possible ipvsadm crash

yangxingwu:


tools/keepalived/keepalived/libipvs-2.6/libipvs.c
```diff
@@ -673,10 +672,9 @@ struct ip_vs_get_dests *ipvs_get_dests(ipvs_service_entry_t *svc)
 	snprintf(dk->iifname, sizeof(dk->iifname), "%s", svc->iifname);
 	snprintf(dk->oifname, sizeof(dk->oifname), "%s", svc->oifname);
 
-	if (dpvs_getsockopt(DPVS_SO_GET_DESTS, dk, len, (void **)&dk_rcv, &len_rcv) < 0) {
+	if (dpvs_getsockopt(DPVS_SO_GET_DESTS, dk, len, (void **)&dk_rcv, &len_rcv)) {
 		free(d);
 		free(dk);
-        dpvs_sockopt_msg_free(dk_rcv);
```
ywc689: The memory pointed by function param `out`  of  `dpvs_getsockopt` can be allocated even if `dpvs_getsockopt` failed. 
Actually, `out` is set NULL if not allocated. I do not think it's the crash reason. 
Similar case is in the  following `ipvs_get_service`.

PR #325, opened 2018-11-07
fix bug: add bklst fail when snat mode.

weiyanhua100:
uint32_t inet_sockaddrip4(struct sockaddr_storage *addr)
在snat场景下，由于其返回类型为uint32_t导致其返回的ip值为0xffffffff，进而导致配置黑名单失败


tools/keepalived/keepalived/check/ipvswrapper.c
```diff
@@ -932,6 +932,7 @@ ipvs_blklst_vsg_cmd(int cmd, list vs_group, virtual_server_t * vs, blklst_addr_g
 static int
 ipvs_blklst_cmd(int cmd, list vs_group, virtual_server_t * vs)
 {
+        uint32_t ip_addr = 0;
```
ywc689: It's better to define `ip_addr` locally where it is used, because it is not used in other cases.

PR #328, opened 2018-11-09
ipvs: add nat-mode support in multi-cores node.

andrewhit:
During the system boots up, the below resource is pre-allcoated.

- Per-socket based connection redirect cache;
- Per-socket based global connection redirect hash table;
- Each lcore allocates the respective packet redirect rings for each other

o When a connection is created and hashed in nat-mode, the related redirect
  is allocated and hashed accordingly.

o When a connection expires to be unhashed and freed in nat-mode, the related
  redirect is unhashed and freed accordingly.

  then check if it matches any connection redirect entry. If matched, enqueue
  the packet into the packet redirect ring of the rediret owner core; otherwise,

  the packet redirect rings owned by the current lcore and process them
  accordingly.

### DIFF
--- a/include/ipvs/conn.h
+++ b/include/ipvs/conn.h
@@ -27,6 +27,11 @@
 #include "ipvs/conn.h"
 #include "ipvs/proto.h"
 #include "ipvs/service.h"
+#include "ipvs/redirect.h"
+
+#define DPVS_CONN_TAB_BITS   20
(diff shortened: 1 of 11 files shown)


andrewhit:
Replaced by https://github.com/iqiyi/dpvs/pull/366

So I am closing this request.

PR #336, opened 2018-11-27
fix some bugs in src/ctrl.c

yangxingwu:


src/ctrl.c
```diff
@@ -538,6 +538,7 @@ int msg_master_process(void)
                     __func__, msg->type);
             add_msg_flags(msg, DPVS_MSG_F_STATE_DROP);
             msg_destroy(&msg);
+            msg_type_put(msg_type);
```
ywc689: `msg_type` is NULL here.

yangxingwu: fixed

ywc689:
DPVS MSG module is updated, and the PR is outdated. Close.

PR #338, opened 2018-11-29
syn proxy: added window size close and update

ytwang0320:
In case of it consumes too many mbufs if RS response is slow.

### DIFF
--- a/src/ipvs/ip_vs_synproxy.c
+++ b/src/ipvs/ip_vs_synproxy.c
@@ -1134,11 +1134,16 @@ static int syn_proxy_send_window_update(int af, struct rte_mbuf *mbuf, struct dp
     }
 
     ack_th = (struct tcphdr *)rte_pktmbuf_prepend(ack_mbuf, sizeof(struct tcphdr));
+    if (!ack_th) {
+        rte_pktmbuf_free(ack_mbuf);
+        RTE_LOG(WARNING, IPVS, "%s:%s\n", __func__, dpvs_strerror(EDPVS_NOROOM));
+        return EDPVS_NOROOM;
+    }
 
     /* Set up tcp header */
     memcpy(ack_th, th, sizeof(struct tcphdr));
     /* clear SYN flag */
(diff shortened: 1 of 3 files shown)


src/ipvs/ip_vs_synproxy.c
```diff
@@ -1109,9 +1110,78 @@ void dp_vs_synproxy_dnat_handler(struct tcphdr *tcph, struct dp_vs_seq *sp_seq)
...
+    }
+
+    ack_th = (struct tcphdr *)rte_pktmbuf_prepend(ack_mbuf, sizeof(struct tcphdr));
+
+    /* Set up tcp header */
+    memcpy(ack_th, th, sizeof(struct tcphdr));
+    /* clear SYN flag */
+    ((uint8_t *)ack_th)[13] &= ~TCP_SYN_FLAG;
+    /* add one to seq and seq will be adjust later */
+    ack_th->seq = htonl(ntohl(ack_th->seq)+1);
```
ywc689: Why should add 1 to seq ? Does it affect subsequent tcp seq adjust?

src/ipvs/ip_vs_synproxy.c
```diff
@@ -1186,6 +1256,14 @@ int dp_vs_synproxy_synack_rcv(struct rte_mbuf *mbuf, struct dp_vs_conn *cp,
...
             return 0;
         }
+	
+        /* Window size has been set to zero in the syn-ack packet to Client.
+	 * If get more than one ack packet here, 
+         * it means client has sent a window probe after one RTO.
+         * The probe will be forward to RS and RS will respond a window update.
+         */
+        if (cp->ack_num == 1)
+            syn_proxy_send_window_update(af, mbuf, cp, pp, th);
```
ywc689: 1. Is it possible `cp->ack_num` is not 1 when retransmission of client's ack?
2. What if the window update packet gets lost? Could we retransmit the lost packet in this case?

PR #341, opened 2018-12-10
fix ipv6 tcp rst bug

mscbg:


src/ipvs/ip_vs_proto_tcp.c
```diff
@@ -1162,11 +1169,14 @@ static int tcp_send_rst(struct dp_vs_proto *proto,
...
             ip6h->ip6_nxt   = IPPROTO_TCP;
-            ip6h->ip6_src  = conn->daddr.in6;
-            ip6h->ip6_dst  = conn->laddr.in6;
+            ip6h->ip6_src   = conn->daddr.in6;
+            ip6h->ip6_dst   = conn->laddr.in6;
+
+            tcp6_send_csum((struct ipv6_hdr *)ip6h, th);
 
-            tcp6_send_csum((struct ipv6_hdr *)ip6h, th); 
+            mbuf->l3_len = sizeof(*ip6h);
```
ywc689: I think `mbuf->l3_len` should be set before checksum calculation.

PR #348, opened 2018-12-18
set errorno after pthread_get/setaffinity_np fails

kldeng:
the return value of pthread_getaffinity_np and pthread_setaffinity_np must be set to the variable errno to describe what went wrong.


mscbg:
@kldeng  please push to devel branch

PR #350, opened 2018-12-21
change arp response from route to addr and del route by refcnt

mscbg:


src/route.c
```diff
@@ -277,22 +246,42 @@ static int route_add_lcore(struct in_addr* dest,uint8_t netmask, uint32_t flag,
...
     if(flag & RTF_LOCALIN || (flag & RTF_KNI)){
         route = route_local_lookup(dest->s_addr, port);
-        error = route_local_del(route);
-        return error;
+        if (!route)
+            return EDPVS_NOTEXIST;
+        list_del(&route->list);
+        rte_atomic32_dec(&route->refcnt);
+        rte_atomic32_dec(&this_num_routes);
+        route4_put(route);
```
ywc689: Is it better to decrease `this_num_routes`  when route is freed?

mscbg: this_num_routes just read routes in list. When route is delete from list, this_num_routes should be decreased, or this_num_routes will not be correct.

PR #355, opened 2018-12-25
1.fix memory leak when getting serivce  which not exsits 2.fix service refcnt incorrect when getting serivce failed

miner-realm:
1.fix memory leak when getting serivce which not exsits
 2.fix service refcnt incorrect when getting serivce failed


ywc689:
@mscbg

PR #371, opened 2019-01-11
ipvs: add connection redirect support in fnat/snat/nat modes.

andrewhit:
Since IPv6 flow director is not well-supported by ixgbe driver and only one
single local IPv6 address can be set in "signature" mode, the below solution of
connection redirect provides the way to support IPv6 packet processing with
multiple local IPv6 addresses configured in "signature" mode.

o Add the switch to enable or disable connection redirect.

o In the case connection redirect is enabled, during the system boots up, the
  below resource is pre-allcoated.
- Per-socket based connection redirect cache;
- Per-socket based global connection redirect hash table;
- Each lcore allocates its respective packet redirect rings for each other

o When a connection is created and hashed in fnat/nat/snat modes, the related
  redirect is allocated and hashed accordingly.

o When a connection expires to be unhashed and freed in fnat/nat/snat modes, the
  related redirect is unhashed and freed accordingly.

  then check if it matches any connection redirect entry. If matched, enqueue
  the packet into the packet redirect ring of the rediret owner core; otherwise,

  the packet redirect rings owned by the current lcore and process them
  accordingly.

### DIFF
--- a/conf/dpvs.conf.items
+++ b/conf/dpvs.conf.items
@@ -11,7 +11,7 @@
 
 ! global config
 global_defs {
-    #daemon                         <disalbe> 
+    #daemon                         <disalbe>
     log_level   INFO                <none>
     log_file    /var/log/dpvs.log   <none>
 }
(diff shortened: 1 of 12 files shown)


jiangsuwwj:
对于nat多核转发，这里还有一种实现思路，不知道是否可行？
inbound的数据包只根据cip + cport + proto 进行hash，
outbound的数据包只根据dip + dport + proto进行hash，
这样数据包回来的时候是不是就落在同一core上避免table miss了呢？

PR #384, opened 2019-02-12
fix bug:

mscbg:
icmp mbuf relate to conn adj but not prepend

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -592,7 +592,7 @@ int ipv4_xmit(struct rte_mbuf *mbuf, const struct flow4 *fl4)
 
     /* output route decision: out-dev, source address, ... */
     rt = route4_output(fl4);
-    if (!rt) {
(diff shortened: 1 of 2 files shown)


src/ipv4.c
```diff
@@ -592,7 +592,7 @@ int ipv4_xmit(struct rte_mbuf *mbuf, const struct flow4 *fl4)
 
     /* output route decision: out-dev, source address, ... */
     rt = route4_output(fl4);
-    if (!rt) {
+    if (!rt || !(rt->flag & RTF_FORWARD)) {
```
ywc689: The following `IP4_INC_STATS(outnoroutes)` is not proper in the case of `!(rt->flag & RTF_FORWARD)` .

mscbg: Yes. But only forward route can be xmit in dpvs, so outnoroutes is OK here

PR #391, opened 2019-02-19
add missing src port in toa module

Danielybl:
When real server creates an IPv6 socket and listens on a port with any address, it can handle IPv4 packet. In this situation, TOA module can still retrieve client's src port. This patch fix the missing for adding client's src port to toa_ip6_data structure.


AdamYYan:
@Danielybl please rebase the new code. now code conflict with the  dpvs/dev

Danielybl:
Hi @ywc689 ,
Sorry for my careless. This fix has an issue. Please revert it fix and I will send another one.

PR #392, opened 2019-02-20
UOA: fix IPv6 UOA issue

Danielybl:
Change this according to test result.

### DIFF
--- a/kmod/uoa/uoa.c
+++ b/kmod/uoa/uoa.c
@@ -803,15 +803,24 @@ static struct uoa_map *uoa_opp_rcv(__be16 af, void *iph, struct sk_buff *skb)
         /* re-calc checksum */
         ip_send_check(iph);
     } else {
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(3,0,0)
-        /* do as upper ipv4, handle for old kernel version */
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4,14,0)
+        /* IPv6 uoa is verified on kernel 4.14.17+ and later.
+         * There is kernel crash for kernel version which is <= 4.9.
+         * The crash is caused by memmove() below which corrupts the pointer
+         * of skb shared info specified by skb->end. The root cause of corrupting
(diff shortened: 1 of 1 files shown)


kmod/uoa/uoa.c
```diff
@@ -803,15 +803,24 @@ static struct uoa_map *uoa_opp_rcv(__be16 af, void *iph, struct sk_buff *skb)
...
     } else {
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(3,0,0)
-        /* do as upper ipv4, handle for old kernel version */
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4,14,0)
+        /* IPv6 uoa is verified on kernel 4.14.17+ and later.
+         * There is kernel crash for kernel version which is <= 4.9.
+         * The crash is caused by memmove() below which corrupts the pointer
+         * of skb shared info specified by skb->end. The root cause of corrupting
+         * is that skb->end is always 320 no mater how long the packet is.
+         * It should be a kernel bug and 4.14 should fix it. */
```
ywc689: Does it crash every time we arrive here? I've tested the uoa.ko on kernel 3.10.0-327 and 2.6.32-696 by UDP IPv6 Fullnat forwarding for some time, and no crash found.

Danielybl: Yes, I tested on kernel version 3.16.0-4 and 4.9.0-0.bpo.8. From the code logic, 2.6.32-696 kernel version will not call memmove(). Does the function works on 2.6.32-696? For 3.10.0-327, I can have a try.

ywc689: For kernel 2.6.32-696, memmove() does not called, but `uoa.ko` does work according to my test.

ywc689:
Uoa passed test on Centos8 (kernel 4.18.0) , Centos7(kernel 3.10.0) and Centos6(kernel 2.6.32)，no crash problem was found. I'm closing this pr. If any problem found later, you can reopen it.

PR #394, opened 2019-02-21
DPVS changes for DPDK 18.11.0

Danielybl:
Some initial changes for DPDK 18.11.0. Compared to DPDK 17.11.2, changes are listed below:
1. API change: rte_eth_dev_count() -> rte_eth_dev_count_avail().
2. In DPDK 18.08 release, ethdev: The old offload API is removed:
       Rx per-port rte_eth_conf.rxmode.[bit-fields]
       Tx per-queue rte_eth_txconf.txq_flags
       ETH_TXQ_FLAGS_NO*
   The transition bits are removed:
       rte_eth_conf.rxmode.ignore_offload_bitfield
       ETH_TXQ_FLAGS_IGNORE
3. In DPDK 18.05, rss_hf check is added. So adapt_device_rss_hf() is added.
   https://github.com/DPDK/dpdk/commit/8863a1fbfc66f762f70eb486431c23cc55f2cc3c
4. The following warning is presented in DPVS log.
   "Kni: kni_add_dev: fail to set mac 6C:92:BF:A1:ED:20 for eth0: Timer expired"
   It doesn't matter as function works well. This log printed because compared to DPDK 17.11.2,
   kni_net_set_mac callback function in DPDK 18.11 adds kni_net_process_request().
   kni_net_process_request() calls wait_event_interruptible_timeout() which returns Timeout err code.
5. When install rte_kni kernel module, parameter "carrier=on" should be added.
   This is because DPDK 18.11 updated the KNI kernel module with a new kernel module parameter,
   carrier=[on|off] to allow the user to control the default carrier state of the KNI kernel network interfaces.
   Actually, rte_kni_update_link API should be used, but it's an experimental API. See release note for detail.

### DIFF
--- a/src/kni.c
+++ b/src/kni.c
@@ -50,6 +50,8 @@ static void kni_fill_conf(const struct netif_port *dev, const char *ifname,
                           struct rte_kni_conf *conf)
 {
     struct rte_eth_dev_info info = {0};
+    const struct rte_bus *bus = NULL;
+    const struct rte_pci_device *pci_dev;
 
     memset(conf, 0, sizeof(*conf));
     conf->group_id = dev->id;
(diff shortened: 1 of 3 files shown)


ywc689:
Is this change compatible with DPDK-17.11.2 ?

Danielybl:
Hi @ywc689 
This change can only works on DPDK-17.11.2. DPDK 18.11.0 has several API change, do we still need to check the DPDK version and adapt the change?

Thanks

ywc689:
The PR is adapted into PR #451 , and would be adopted later.

PR #395, opened 2019-02-21
sa_pool: add a sa_pool option using laddr to map lcore except the lport-mask mode

kldeng:
sa_pool: add a sa_pool option using laddr to map lcore except the lport-mask mode

NOTICE:
number of laddrs should be greater or equal to number of slave lcores.
Or, some slave lcores will have no laddr causing FNAT fowarding failed in
those slave lcores.

### DIFF
--- a/src/sa_pool.c
+++ b/src/sa_pool.c
@@ -69,6 +69,9 @@
 #define SAPOOL_MIN_HASH_SZ  1
 #define SAPOOL_MAX_HASH_SZ  128
 
+#define LPORT_LCORE_MAPPING_POOL_MODE_NAME "lport_lcore_mapping"
+#define LADDR_LCORE_MAPPING_POOL_MODE_NAME "laddr_lcore_mapping"
+
 enum {
     SA_F_USED               = 0x01,
 };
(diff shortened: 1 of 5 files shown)


kldeng:
Since some 10GbE controller like 82599 only support signature mode for IPv6. And it was found that only one IPv6 address can be supported in RTE_ETH_FLOW_NONFRAG_IPV6_TCP flow_type, but multiple IPv6 addresses can be supported in RTE_ETH_FLOW_NONFRAG_IPV6_OTHER. 
So, we can streer the incoming IPv6 packets to the correct lcore by laddr_lcore_mapping mode. This patch can be another solution except commit a9d16e301a48e9 and do no harm to performance.

src/sa_pool.c
```diff
@@ -292,9 +331,51 @@ static int sa_pool_free_hash(struct sa_pool *ap)
...
+        err = EDPVS_NOMEM;
+        goto errout;
+    }
+
+    ap->ifa = ifa;
+    ap->low = low;
+    ap->high = high;
+    rte_atomic32_set(&ap->refcnt, 0);
+
+    err = sa_pool_alloc_hash(ap, sa_pool_hash_size, fdir);
```
Danielybl: Hi,
Do you still need to change this function according to pool_mode? As the some port is masked in lport_lcore_mapping_mode.

src/sa_pool.c
```diff
@@ -270,7 +309,7 @@ static int sa_pool_alloc_hash(struct sa_pool *ap, uint8_t hash_sz,
         for (port = ap->low; port <= ap->high; port++) {
             struct sa_entry *sa;
 
-            if (fdir->mask &&
+            if (sa_pool_mode == LPORT_LCORE_MAPPING_POOL_MODE && fdir->mask &&
```
kldeng: Hi, @Danielybl  It's done here.

PR #406, opened 2019-03-08
Update ipv4_frag.c, mismatch of rte_socket_id rte_lcore_id

icedustdream:
mismatch of rte_lcore_id and rte_socket_id


ywc689:
The patch is already merged into devel branch.

PR #413, opened 2019-03-19
force remove inhibited real servers

yangxingwu:


ytwang0320:
> @ytwang0320

approved +1

PR #415, opened 2019-03-20
ndisc/neigh: sync the neighbour entry to other cores once it is created.

andrewhit:
During nat64 test, the below issue is met on IPv6-side (aka wan side).
Provided that dpvs's gateway IPv6 address is  fcbd:dc00:46::1.

1) When the outgoing packet destined to IPv6 client is sent out from dpvs, we
  need to resolve dpvs's gateway mac address. So the below neighbour
  solicitation is sent from dpvs to its IPv6 gateway.

  (:: -> ff02::1:ff00:1) that includes info, "Neighbour Solicitation for
  fcbd:dc00:46::1".

  And the related neighbour entry, E1, (fcbd:dc00:46::1, null mac address, dpdk1) is
  created on core A.

2) The coresponding neighbour advertisement is received on core B.

  (fe80:274:9cff:fe0e:1441 -> ff02::1) that includes info, "Neighbour

  Since it is received on core B, we have no way to find the previously created
  neighbour entry, E1 which is created on core A.

  So the solution is to sync the neighbour entry to the other cores once it is
  created and we still follow the design of per-lcore neighbour mgmt.

  Additionally, within ndisc_recv_na(), target address, "fcbd:dc00:46::1" should
  be used to find the neighbour entry instead of source-ip address,
  "fe80:274:9cff:fe0e:1441". So the entry, E1 is resolved as below and then
  synced to the other cores.

      (fcbd:dc00:46::1, 00:74:9c:0e:14:41, dpdk1)

### DIFF
--- a/src/neigh.c
+++ b/src/neigh.c
@@ -595,10 +595,9 @@ int neigh_output(int af, union inet_addr *nexhop,
                 neigh_entry_state_trans(neighbour, 0);
             }
             return EDPVS_OK;
-        }
-        else if ((neighbour->state == DPVS_NUD_S_REACHABLE) ||
-                 (neighbour->state == DPVS_NUD_S_PROBE) ||
-                 (neighbour->state == DPVS_NUD_S_DELAY)) {
+        } else if ((neighbour->state == DPVS_NUD_S_REACHABLE) ||
+                   (neighbour->state == DPVS_NUD_S_PROBE) ||
+                   (neighbour->state == DPVS_NUD_S_DELAY)) {
 
             neigh_fill_mac(neighbour, m, NULL, port);
             netif_xmit(m, neighbour->port);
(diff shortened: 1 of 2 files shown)


src/neigh.c
```diff
@@ -639,6 +640,15 @@ int neigh_output(int af, union inet_addr *nexhop,
...
 
+        /*
+         * Sync to other cores though it has not been resolved. And the reason
+         * is that neighbour solicitation is sent from core A where the related
+         * neighbour entry is created. However the related neighbour
+         * advertisement is received on core B. So if not synced, we have no
+         * chance to resolve it.
+         */
+        neigh_sync_core(neighbour, 1, NEIGH_ENTRY);
+
```
mscbg: It's OK for core A send ns and core B received na. 'ndisc_recv_ns' will create entry in core B and do 'neigh_sync_core' . Did you have some problem with it?

andrewhit: It works fine and I will raise a new pull request.

andrewhit:
Replaced with https://github.com/iqiyi/dpvs/pull/418

So I am closing it.

PR #418, opened 2019-03-26
ndisc/neigh: sync the neighbour entry to other cores once it is created.

andrewhit:
During nat64 test, the below issue is met on IPv6-side (aka wan side).
Provided that dpvs's gateway IPv6 address is  fcbd:dc00:46::1.

1) When the outgoing packet destined to IPv6 client is sent out from dpvs, we
   need to resolve dpvs's gateway mac address. So the below neighbour
   solicitation is sent from dpvs to its IPv6 gateway.

   (:: -> ff02::1:ff00:1) that includes info, "Neighbour Solicitation for
   fcbd:dc00:46::1".

   And the related neighbour entry, E1, is created on core A.

        E1: (fcbd:dc00:46::1, 00:00:00:00:00:00, dpdk1, DPVS_NUD_S_SEND)

2) The coresponding neighbour advertisement is received on core B.

   (fe80:274:9cff:fe0e:1441 -> ff02::1) that includes info, "Neighbour

   Also within ndisc_recv_na(), target address, "fcbd:dc00:46::1" should
   be used to find the neighbour entry instead of source-ip address,
   "fe80:274:9cff:fe0e:1441".

   Since the neighbour entry is not found on core B, we just create the below
   entry, E2 directly with the state of "DPVS_NUD_S_REACHABLE" and then sync it
   to the other cores. Lastly, the entry, E1 created on core A gets resolved
   accordingly.

        E2: (fcbd:dc00:46::1, 00:74:9c:0e:14:41, dpdk1, DPVS_NUD_S_REACHABLE)

### DIFF
--- a/src/ipv6/ndisc.c
+++ b/src/ipv6/ndisc.c
@@ -414,10 +414,7 @@ static int ndisc_recv_na(struct rte_mbuf *mbuf, struct netif_port *dev)
     struct neighbour_entry *neigh;
     struct inet_ifaddr *ifa;
     int hashkey;
-
-    struct in6_addr *saddr = &((struct ip6_hdr *)mbuf->userdata)->ip6_src;
     struct in6_addr *daddr = &((struct ip6_hdr *)mbuf->userdata)->ip6_dst;
-
     struct nd_msg *msg = rte_pktmbuf_mtod(mbuf, struct nd_msg *);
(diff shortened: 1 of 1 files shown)


andrewhit:
@mscbg  please have a review on the new patch based on your comments.

PR #420, opened 2019-03-26
ipv6: pass into kni the icmpv6 echo reqeust packets destined to vip6.

andrewhit:
o Within nat64 case, the below route is configured within keepalived.conf.

```
  static_routes {
      # wan-side IPv6 gateway
      ::/0 via fcbd:dc00:46::1 dev dpdk1

      # lan-side IPv4 routes
      . . .
  }
```

  When the client ping6 vip6, like fcbd:dc01:1:111::2 configured on dpvs, the
  packet matches the default IPv6 route and then gets dropped becuase IPv6
  forwarding is disabled. So the client fails to ping6 vip6.

  The solution is to pass the to-host packet to kni when ipv6 forwarding is
  disabled.

### DIFF
--- a/src/ipv6/ipv6.c
+++ b/src/ipv6/ipv6.c
@@ -489,8 +489,12 @@ static int ip6_rcv_fin(struct rte_mbuf *mbuf)
     if (rt->rt6_flags & RTF_LOCALIN) {
         return ip6_local_in(mbuf);
     } else if (rt->rt6_flags & RTF_FORWARD) {
-        /* pass multi-/broad-cast to kni */
-        if (etype != ETH_PKT_HOST)
+        /*
+         * pass the packet to kni in the below two cases.
+         * - multi-/broad-cast packets;
+         * - to-host packets when ipv6 forwarding is disabled;
+         */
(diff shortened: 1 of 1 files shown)


mscbg:
I don’t understand. If you configured  fcbd:dc01:1:111::2 on dpvs. There should be host route in dpvs, and default route will not be matched.

andrewhit:
In this case,  VIP: fcbd:dc01:1:111::2 is not configured via dpip, but configured in keepalived.conf as below.

```
virtual_server_group vip_fnat {
    fcbd:dc01:1:111::2 80
}
```

Thus no host route (to VIP) is generated and the ICMPv6 packet matches the default route accordingly.

mscbg:
Then I think you should configure VIP with dpip. Think about LVS, LVS configured VIP/port by keepalived/ipvsadm, but also VIP is configured in eth.
By the way, IPv6 address will generate a muticast MAC to be set to eth. And you should better configured VIP in kni. Dpdk0 in dpvs will has the same muticast MAC with dpdk0.kni

andrewhit:
Cancelling this patch for the same reason described in https://github.com/iqiyi/dpvs/pull/425

PR #423, opened 2019-03-28
ipv6: add the debugging info for showing IPv6 header.

andrewhit:


src/ipv6/ipv6.c
```diff
@@ -75,6 +75,22 @@ static RTE_DEFINE_PER_LCORE(struct inet_stats, ip6_stats);
         this_ip6_stats.__f__##octets += (val); \
     } while (0)
 
+#ifdef CONFIG_DPVS_MBUF_DEBUG
+static inline void ip6_show_hdr(const char *func, struct rte_mbuf *mbuf)
```
mscbg: CONFIG_DPVS_MBUF_DEBUG is not a good name, because we didn't dump mbuf but just ipv4/ipv6 header

andrewhit: How about using "CONFIG_DPVS_IP_HEADER_DEBUG"?

mscbg: I think it's fine.

andrewhit: Done.

PR #425, opened 2019-03-28
neighbour: use IPv6 any address as the source IP of neighbour solicit…

andrewhit:
…ation.

o Within the case of "ping6 vip6", we prefer processing ICMPv6 echo request in
  userspace instead of passing this request to kernel stack via kni. So we need to
  configure vip6, like fcbd:dc01:1:111::2 on the interface in either of the
  below ways.

  1. Configuration within keepalived.conf
```
  static_ipaddress {
      fcbd:dc01:1:111::2/128 dev dpdk1
  }
```

  2. Configuration via `  # dpip addr add fcbd:dc01:1:111::2/128 dev dpdk1
`
  And then its related to-host route is generated as below and IPv6 default route displayed via "dpip -6 route show"
`        "inet6 fcbd:dc01:1:111::2/128 dev dpdk1 mtu 1500 scope host"`

o With the above configuration, ICMPv6 echo request like below is processed in
  LOCAL_IN path since its dst-IP (aka vip6) matches the above to-host route.

`        fcbd:dc01:1:333::3 -> fcbd:dc01:1:111::2`

  Then dpvs generates the corresponding ICMPv6 echo reply which matches against
  the IPv6 default route configured as below.

```
  static_routes {
      ::/0 via fcbd:dc00:46::1 dev dpdk1  // IPv6 default route

      10.10.127.66/26 dev dpdk0
      10.10.127.0/26 via 10.10.127.65 src 10.10.127.66 dev dpdk0
  }
```
  - IPv6 default route displayed via "dpip -6 route show"
    "inet6 default via fcbd:dc00:46::1 dev dpdk1 mtu 1500 scope global"

  Then we need to resolve the mac address (aka link layer address) for the
  nexthop (gateway) gotten from the default IPv6 route, fcbd:dc00:46::1 in order
  to send out ICMPv6 echo reply.

  So the below neighbour solicitation is sent out from dpvs.

```
  (fcbd:dc01:1:111::2 -> ff02::1:ff00:1) includes the info, "Neighbour
  Solicitation for fcbd:dc00:46::1 from 90:e2:ba:ea:dc:c0".
```

  However the coresponding neighbour advertisement is never received from
  gateway.

  The root cause is that vip6 is selected as the source IP as neighbour
  solicitation according to the nexthop address, fcbd:dc00:46::1. And the
  solution is to use IPv6 any address as the source IP of neighbour solicitation
  as below and the coresponding neighbour advertisement is received consequently.

```
  Neighbour Soliction:
      (:: -> ff02::1:ff00:1) includes the info, "Neighbour Solicitation for
      fcbd:dc00:46::1".
```

  Neighbour Advertisement:
      (fe80::274:9cff:fe0e:1441 -> ff02::1) includes the info, "Neighbour

### DIFF
--- a/src/neigh.c
+++ b/src/neigh.c
@@ -431,13 +431,6 @@ static void neigh_state_confirm(struct neighbour_entry *neighbour)
             RTE_LOG(ERR, NEIGHBOUR, "[%s] send arp failed\n", __func__);
         }
     } else if (neighbour->af == AF_INET6) {
-        /*to be continue*/
-        ipv6_addr_copy(&daddr.in6, &neighbour->ip_addr.in6);
-        inet_addr_select(AF_INET6, neighbour->port, &daddr, 0, &saddr);
-
-        if (ipv6_addr_any(&saddr.in6))
-            RTE_LOG(ERR, NEIGHBOUR, "[%s]no source ip\n", __func__);
-
         ndisc_solicit(neighbour, &saddr.in6);
     }
 }
@@ -1079,4 +1072,3 @@ int neigh_term(void)
 
     return EDPVS_OK;
 }
-


mscbg:
You should configure route for ipv6. Such as dpip add fcbd:dc01:1:111::2/64 and gateway should be configured in the same net such as  fcbd:dc01:1:111::3. Solicit with any address is DAD，Gateway recieve DAD and find ip conflict，then it will send NA

andrewhit:
Here is the topology of this case.

Screen Shot 2019-04-24 at 2 34 44 PM

VIP configured on  eth2 (aka dpdk1) of DPVS is fcbd:dc01:1:111::2 and its gateway is fcbd:dc00:46::1, for we do not require VIP and its gateway should be in the same network. And it is the typical configuration that exists in the scenarios of IPv4-to-IPv4, NAT64 or IPv6-to-IPv6.

Additionally,  fcbd:dc00:46::2 configured on eth2 of DVPS is used to communicate with its gateway, fcbd:dc00:46::1 configured on the switch via ospf packets sent between DPVS and the switch. The application of bird6 is running on DPVS.

And configuring only ONE single IPv6 route is for the purpose of simplicity in the above NAT64 case.

```
static_routes {
    ::/0 via fcbd:dc00:46::1 dev dpdk1 // IPv6 default route

    10.10.127.66/26 dev dpdk0
    10.10.127.0/26 via 10.10.127.65 src 10.10.127.66 dev dpdk0
}

```

mscbg:
The  topology is almost the same with us. But if you configured  fcbd:dc00:46::2/126 in dpvs.  fcbd:dc00:46::2 will be the src ip to send NS to a muticast address. Never choose any address(::) to send NS, that means DAD. And perhaps you didn't configured VIP in dpvs or kni. IPv6 address should configured then dpvs will add a muticast MAC, or NS will not be recieved in dpvs.(dest ip in NS is a muticast address)

andrewhit:
As @mscbg mentioned, it is a bad idea that dpvs sends out DAD packet in order to resolve its neighbour IPv6 address (aka fcbd:dc00:46::1).

After discussing with @mscbg, vip6 and public IPv6 interface address (aka  fcbd:dc00:46::2) should be configured in the section of static_ipaddress { } within keepalived.conf.  And same to the private interface address.

With this solution, DPVS sends out NS whose source address is  fcbd:dc00:46::2 as the attachment in the user-space.

Screen Shot 2019-04-25 at 8 41 31 PM

Here is the configuration in the section of static_ipaddress { }.
```
static_ipaddress {
    # vip6 address
    fcbd:dc01:1:111::2/128 dev eth2

    # public interface: ospf communication between switch and ttgw
    fcbd:dc00:46::2/126 dev eth2

    # private interface
    10.10.127.66/26 dev eth3
}

static_routes {
    ::/0 via fcbd:dc00:46::1 dev eth2
    10.10.127.0/26 via 10.10.127.65 src 10.10.127.66 dev eth3
}
```

And here are the IPv6 routes generated from above.
```
root@n10-023-013:~/debs/nat64# dpip -6 route show
inet6 fcbd:dc00:46::2/128 dev eth2 mtu 1500 scope host
inet6 fcbd:dc01:1:111::2/128 dev eth2 mtu 1500 scope host
inet6 fcbd:dc00:46::/126 dev eth2 mtu 1500 scope link 
inet6 default via fcbd:dc00:46::1 dev eth2 mtu 1500 scope global
```

Also please note the below line generates two IPv6 routes as below.
```
static_ipaddress {
    # public network: ospf between switch and ttgw
    fcbd:dc00:46::2/126 dev eth2
}

inet6 fcbd:dc00:46::2/128 dev eth2 mtu 1500 scope host
inet6 fcbd:dc00:46::/126 dev eth2 mtu 1500 scope link 
```

So this patch should be abandoned.

PR #436, opened 2019-04-23
tcp_check/ssl_get/http_get/misc_check supports retry

yangxingwu:
The default value is 1 retry after 1 second

please check https://github.com/acassen/keepalived (commit id:
560654294a28c5f14b4da077bfa4e5b6040dd89d) for reference


ywc689:
keepalived is upgraded to 2.0.19, and the problem fixed.

PR #438, opened 2019-04-28
ndisc/neighbor: add the missing saddr and correct the debugging info.

andrewhit:


src/neigh.c
```diff
@@ -558,7 +558,7 @@ static inline void neigh_show_nexthop(const char *func, int af,
 
     RTE_LOG(DEBUG, NEIGHBOUR,
             "%s: [%d] port %s, nexthop %s\n",
-            __func__, rte_lcore_id(), port->name,
+            func, rte_lcore_id(), port->name,
```
ywc689: Why change `__func__` to `func`? Is the later defined?

andrewhit: Use the input parameter, func instead so we can know which function calls neigh_show_nexthop().

PR #439, opened 2019-04-29
changed conn redirect free logic to avoid crash

ytwang0320:


src/ipvs/ip_vs_conn.c
```diff
@@ -106,8 +106,10 @@ static void dp_vs_conn_free(struct dp_vs_conn *conn)
     if (!conn)
         return;
 
-    dp_vs_redirect_free(conn);
-
+    if (conn->redirect) {
```
andrewhit: Better to check conn within dp_vs_redirect_free().

src/ipvs/ip_vs_redirect.c
```diff
@@ -85,14 +85,13 @@ dp_vs_redirect_alloc(enum dpvs_fwd_mode fwdmode)
     return r;
 }
 
-void dp_vs_redirect_free(struct dp_vs_conn *conn)
+void dp_vs_redirect_free(struct dp_vs_redirect *redirect)
 {
-    if (conn->redirect) {
+    if (redirect) {
```
andrewhit: Adding the check against "conn" is enough and no need to expose "conn->redirect = NULL" to outside.

ywc689:
The problem is fixed in latest devel code.

PR #440, opened 2019-05-07
Add support for i40e

Danielybl:
1. x-port https://github.com/DPDK/dpdk/commit/97bd4ef9a0fecfc660231d9547f94a8df616e8df from DPDK 18.02
2. uncomment DPVS_CFG_I40E_PMD

The commit 97bd4ef9a0fecfc660231d9547f94a8df616e8df has been modified as
0003-net-i40e-support-input-set-configuration.patch.


Danielybl:
> It seems incomplete. Also need to add support for fdir mask for each queue.

It's a bit hacking, but it can work. The fdir mask is set on port and no need to configure it on each queue.

ytwang0320:
> > It seems incomplete. Also need to add support for fdir mask for each queue.
> 
> It's a bit hacking, but it can work. The fdir mask is set on port and no need to configure it on each queue.

Hi Daniel,
I have tried this patch using X722 NIC. The response really can hit the core where the request send out. But the traffic for one service always work on one same core and it cannot be well distributed on each core. I feel confused if the input set configuration break RSS logic.

Danielybl:
> > > It seems incomplete. Also need to add support for fdir mask for each queue.
> > 
> > 
> > It's a bit hacking, but it can work. The fdir mask is set on port and no need to configure it on each queue.
> 
> Hi Daniel,
> I have tried this patch using X722 NIC. The response really can hit the core where the request send out. But the traffic for one service always work on one same core and it cannot be well distributed on each core. I feel confused if the input set configuration break RSS logic.

Hi,
I checked the RSS on DPDK18.11.0 and I can distribute. I forgot to see if RSS works on DPDK17.11. I will check it again.

Thanks,
Daniel

oupeng007:
i40e上你们有没有发现运行一段时间高概率软件收不到包，取包i40e_rx_scan_hw_ring函数中判断 I40E_RX_DESC_STATUS_DD_SHIFT位一直为0，但是硬件是收到包了

PR #448, opened 2019-06-17
SVC/DEST/LADDR support percore

mscbg:


src/ipvs/ip_vs_stats.c
```diff
@@ -40,175 +40,20 @@ static void __dp_vs_stats_clear(struct dp_vs_stats *stats)
...
-    *p = alloc_percpu_stats();
-    if (NULL == *p) {
-        RTE_LOG(WARNING, SERVICE, "%s: no memory!\n", __func__);
-        return EDPVS_NOMEM;
-    }
+    dst->conns    += src->conns;
+    dst->inpkts    += src->inpkts;
+    dst->inbytes  += src->inbytes;
+    dst->outbytes += src->outbytes;
+    dst->outpkts  += src->outpkts;
```
ywc689: ajust code alignment here

tools/ipvsadm/ipvsadm.c
```diff
@@ -200,6 +200,7 @@ static const char* optnames[] = {
 	"ifname" ,
 	"sockpair" ,
 	"hash-target",
+        "cpu"
```
ywc689: fix indentation

tools/ipvsadm/ipvsadm.c
```diff
@@ -1810,6 +1864,9 @@ print_service_entry(ipvs_service_entry_t *se, unsigned int format)
 			left -= snprintf(svc_name + strlen(svc_name), left,
 				"MATCH %s", proto);
 
+			left -= snprintf(svc_name + strlen(svc_name), left,
+				",af=%s", af);
+
```
ywc689: "af" can be induced from the IP address in the "srange" and "drange". Do we need to display it? 
If "af" should be displayed here, I think it's better to place the string before "proto" and remove the preceding "af=" .

src/ipvs/ip_vs_laddr.c
```diff
@@ -485,73 +476,200 @@ static int laddr_sockopt_set(sockoptid_t opt, const void *conf, size_t size)
...
+static int dp_vs_copy_percore_laddrs_stats(struct dp_vs_laddr_conf *master_laddrs,
+                                           struct dp_vs_laddr_conf *slave_laddrs)
+{
+    int i;
+    if (master_laddrs->nladdrs != slave_laddrs->nladdrs)
+        return EDPVS_INVAL;
+    for (i = 0; i < master_laddrs->nladdrs; i++) {
+        master_laddrs->laddrs[i].nport_conflict += slave_laddrs->laddrs[i].nport_conflict;
+        master_laddrs->laddrs[i].nconns += slave_laddrs->laddrs[i].nconns;
     }
```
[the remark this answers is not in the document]
mscbg: If not， master_laddrs->nladdrs will not equal to slave_laddrs->nladdrs.

tools/ipvsadm/ipvsadm.c
mscbg: What if there is no range.

PR #453, opened 2019-07-03
ipvs: decrease the service's usecnt before deleting the service.

andrewhit:


andrewhit:
add/edit/delete service checks DPVS_WAIT_WHILE(rte_atmomic32_read(&svc->usecnt) > 1).

With the above fix, we need to change the above line into DPVS_WAIT_WHILE(rte_atmomic32_read(&svc->usecnt) > 0) within dp_vs_del_service().

And it is not good so closing this request.

PR #458, opened 2019-07-24
add dpvs traffic stats

xiaguiwu:


src/iftraf.c
```diff
@@ -0,0 +1,1051 @@
...
+
+    if (likely(iftraf_disable)) {		
+        return;
+    }	
+
+    ticket++;
+    if(ticket % TIME_INTERVAL == 0) {
+        history_rotate();
+        ticket = 0;
+    }
```
ywc689: 1. "history_rotate" is called less frequently if traffic is heavy, so we cannot get the topN traffic in Bps or bps because no time info is recorded.  If we remember the timestamp or cpu ticks every time we call "history_rotate", can we figure out a solution to acquire the approximation of the topN traffic per second?
2. How long does "history_rotate" take when traffic is heavy? Would it starve other jobs in this CPU?

src/iftraf.c
```diff
@@ -0,0 +1,1051 @@
...
+    void* data;
+} sorted_list_node;
+
+typedef struct {
+    sorted_list_node root;
+    int (*compare)(void*, void*);
+} sorted_list_type;
+
+int sorted_list_num = 0;
+sorted_list_type iftraf_sorted_list;
```
ywc689: 1. Place Macro "TIME_INTERVAL" before variable definitions.
2. Use "static" global variables if they are used in this file only.
3. Use module name prefix for global naming to reduce conflict chances, such as: "ticket"->"iftraf_ticket", "TIME_INTERVAL"->"IFTRAF_TIME_INTERVAL", ...

PR #475, opened 2019-09-05
support for trace route 

ytwang0320:
icmp_send can only work when pkt's dest is our's, so traceroute is not work.

### DIFF
--- a/src/ipset.c
+++ b/src/ipset.c
@@ -337,8 +337,7 @@ static struct dpvs_sockopts ipset_sockopts = {
 
 static int ipset_parse_conf_file(void)
 {
-    char *buf, *pstr;
-    bool found_num = false;
+    char *buf, ch;
     struct dp_vs_multi_ipset_conf *ips = NULL;
     int ip_num = 0, ipset_size = 0, ip_index = 0;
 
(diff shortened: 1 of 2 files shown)


src/ipv4.c
```diff
@@ -527,7 +527,7 @@ int ipv4_xmit(struct rte_mbuf *mbuf, const struct flow4 *fl4)
     struct route_entry *rt;
     struct ipv4_hdr *iph;
 
-    if (!mbuf || !fl4 || fl4->fl4_saddr.s_addr == htonl(INADDR_ANY)) {
+    if (!mbuf || !fl4) {
```
ywc689: Why  remove source check? Does "traceroute" work after removing it?
I think the following aspects should be considered for traceroute:
1. traceroute should support "virtual ip or host ip" configured in dpvs, some icmp types, such as ICMP_DEST_EXCEEDED, ICMP_DEST_UNREACH, may need to construct.
2. traceroute shoud support ipvs snat mode, dpvs should support the forwarding of traceroute udp/icmp packets.
3. traceroute should support both ipv4 and ipv6.
2. traceroute should

src/ipset.c
```diff
@@ -347,24 +346,32 @@ static int ipset_parse_conf_file(void)
         RTE_LOG(WARNING, IPSET, "no memory for ipset buf\n");
         return -1;
     }
+
+    while(!feof(g_current_stream))
+    {
+        if((ch=getc(g_current_stream))=='\n')
```
ywc689: Do we need to remove the blank lines? We may use a  "readline" function(such as "readn" in src/common.c) to get this.

src/ipv4.c
donghaobo: > Why remove source check? 

maybe the question is why check source address here?  in this function,  line 654   `if (iph->src_addr == htonl(INADDR_ANY)) {` is checking src_addr eq ANY, and will select an address on the port as src_addr.
this function is called only at `icmp_echo` and `icmp_send`.
`icmp_echo` is for `ICMP_ECHO` processing. it swaped `src_addr` & `dst_addr`, then called `ipv4_xmit`. this change does nothing different.
`icmp_send` is used for generating icmp packets. if `dst_addr` is local address, the out icmp packet will use it as source address. if not, should select an local address for the out packet. just set `src_addr` `ANY`, and `ipv4_xmit` will do the selection. 
the selection first get the out interface by `route4_output`, and then use `inet_addr_select` to get a local address on that port.
@ywc689

donghaobo: > > Why remove source check?
> 
> maybe the question is why check source address here? in this function, line 654 `if (iph->src_addr == htonl(INADDR_ANY)) {` is checking src_addr eq ANY, and will select an address on the port as src_addr.
> this function is called only at `icmp_echo` and `icmp_send`.
> `icmp_echo` is for `ICMP_ECHO` processing. it swaped `src_addr` & `dst_addr`, then called `ipv4_xmit`. this change does nothing different.
> `icmp_send` is used for generating icmp packets. if `dst_addr` is local address, the out icmp packet will use it as source address. if not, should select an local address for the out packet. just set `src_addr` `ANY`, and `ipv4_xmit` will do the selection.
> the selection first get the out interface by `route4_output`, and then use `inet_addr_select` to get a local address on that port.
> @ywc689

this is used for ipvs snat mode.

ywc689:
Adopt traceroute solution of #657 #630.
Ipset for gfwip should adjust codes. Just close it now. Reopen it if any update later.

PR #480, opened 2019-10-11
alloc_percpu_stats: alloc memory from all sockets

yangxingwu:
in most cases, this function run on master core and socket 0


ywc689:
The ipvs stats module is changed in pr #448 , which we will merged soon.

PR #501, opened 2019-11-26
fdir: add "filter" switch of fdir config to support multi-core processing in VM

andrewhit:
o By default, fdir filter is always enabled.
```
netif_defs {
    ...
    <init> device dpdk0 {
        fdir {
            ...
            <init> filter  on
        }
    }
    ...
}
```

o In some case that dpvs is deployed in some VM, VM does not support fdir filter
  configuration so only ONE single worker core can be used accordingly. In fact,
  we don't need to add fdir filters, but create only the specific sa pools.

  With "packet redirect" enabled, multiple worker cores can be used accordingly
  for the benefit of performance, like cps/pps.

```
netif_defs {
    ...
    <init> device dpdk0 {
        fdir {
            filter  off
        }
    }
    ...
}

ipvs_defs {
      conn {
          redirect  on
      }
  }
```

### DIFF
--- a/conf/dpvs.conf.items
+++ b/conf/dpvs.conf.items
@@ -34,6 +34,7 @@ netif_defs {
             descriptor_number   512         <512, 16-8192>
         }
         fdir {
+            <init> filter       on          <on, on/off>
             mode                perfect     <perfect, none|signature|perfect|perfect_mac_vlan|perfect_tunnel>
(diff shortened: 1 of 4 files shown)


andrewhit:
> Please add the new config entry "filter" into file "conf/dpvs.conf.items"

Done as requested.

PR #505, opened 2019-12-01
ipvs: remove unnecessary check for tunnel termination

gogoxiaoxiao:
Since ip_tunnel_term_tab always return EDPVS_OK, it's unecessary to
check the return code.

Signed-off-by: Haishuang Yan <yanhaishuang@cmss.chinamobile.com>


ywc689:
I do not think it's a problem.  Just a coding style.

PR #532, opened 2020-02-17
[WIP] Replace black list to white list.

haosdent:
- [x] Rename blklst to whtlst.
- [x] Rename BLKLST to WHTLST.
- [x] Rename black to white.
- [x] Increase CMD numbers for whitelist to avoid conflict with blacklist.
- [x] Chnage shortcut CMD from -B to -W, -U to -O, -V to -Y


haosdent:
Would post manual test result at here at this Sat.

ywc689:
Blacklist, whitelist are different safety strategies. Someone may have already been using blacklist now. I think we should keep both.

haosdent:
Close in favor of https://github.com/iqiyi/dpvs/pull/560

PR #554, opened 2020-04-11
toa: compatiable with kernel version 4.17+

everpcpc:
tested with 5.4.28


ywc689:
Compile failed on Centos8.1 (kernel: 4.18.0-147). 
PR #571 also provided a TOA kmod bugfix of high kernel version.  It's almost the same as your pr, but it covered more kernel versions. So PR #571 would be accepted instead of this PR.

PR #556, opened 2020-04-13
Retrieve dev from `conn->in_dev` for inbound traffic when fast_xmit is enabled

haosdent:
At 1.6, `conn->in_dev` is setted when sending checkshum https://github.com/iqiyi/dpvs/blob/DPVS-1.6-LTS/src/ipvs/ip_vs_proto_tcp.c#L591-L594

```
    if (rt && rt->port)
        dev = rt->port;
    else if (conn->in_dev)
        dev = conn->in_dev;
```

However, this is removed in https://github.com/iqiyi/dpvs/commit/e6947f826cc05689d3d901bd794fc0cd0f96802d

This pull request try to fix this issue to ensure fast_xmit works correctly.

### DIFF
--- a/src/ipvs/ip_vs_proto_tcp.c
+++ b/src/ipvs/ip_vs_proto_tcp.c
@@ -168,6 +168,8 @@ static inline int tcp_send_csum(int af, int iphdrlen, struct tcphdr *th,
         struct ip6_hdr *ip6h = ip6_hdr(mbuf);
         if (rt6 && rt6->rt6_dev)
             dev = rt6->rt6_dev;
+        else if (conn->in_dev)
+            dev = conn->in_dev;
         else if (conn->out_dev)
             dev = conn->out_dev;
         if (likely(dev && (dev->flag & NETIF_PORT_FLAG_TX_TCP_CSUM_OFFLOAD))) {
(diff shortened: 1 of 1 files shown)


src/ipvs/ip_vs_proto_tcp.c
```diff
@@ -185,6 +187,8 @@ static inline int tcp_send_csum(int af, int iphdrlen, struct tcphdr *th,
         struct ipv4_hdr *iph = ip4_hdr(mbuf);
         if (rt && rt->port)
             dev = rt->port;
+        else if (conn->in_dev)
+            dev = conn->in_dev;
```
ywc689, line 191: I think it doesn't make sense to using `in_dev` when sending pakets out.

haosdent: I see. Need to add a new parameter to indentify send or receive?

ywc689: Yes. You can specify the direction (INBOUND/OUTBOUND) with a function parameter.

haosdent:
Got it, let me update.

ywc689:
The problem is fixed by #761 , and DPVS-1.7-LTS is out of maintenance now. So close the PR.

PR #560, opened 2020-04-14
Support white list.

haosdent:
Have verified in our test cluster.

```
# ipvsadm -h|grep whitelist
  ipvsadm -O|Y -t|u|q|f service-address -2 whitelist-address
  --add-whtlst    -O            add whitelist address
  --del-whtlst    -Y            del whitelist address
  --get-whtlst    -W            get whitelist address
```

- [x] Pass if not whitelist IPs configured for the specific VIP.
- [x] Reject if whitelist IPs are configured and whitelist IPs are mismatched with the incoming IP for the specific VIP.
- [x] Reject if whitelist IPs are configured after the connection has established and whitelist IPs are mismatched with the incoming IP for the specific VIP.


ywc689:
The feature was updated in pr #691 to adapt to the `devel` branch.
If the PR is to be accepted into DPVS-1.7-LTS, some problems should be fixed. Refer to code reviews of pr #691 for details.

PR #573, opened 2020-05-14
dpvs: fix the problem of  showing the laddr info of  a slave worker

legend050709:
dpvs return the error log "SERVICE: laddr_sockopt_get: find no laddr for cid=2" in the function laddr_sockopt_get because of the different cid  when executing ipvsadm -G --cpu N (N > 0, which is the cpu index of a slave worker).

### DIFF
--- a/src/ipvs/ip_vs_laddr.c
+++ b/src/ipvs/ip_vs_laddr.c
@@ -525,6 +525,7 @@ static int get_msg_cb(struct dpvs_msg *msg)
 
     *laddrs = *laddr_conf;
     laddrs->nladdrs = naddr;
+    laddrs->cid = cid;
     for (i = 0; i < naddr; i++) {
         laddrs->laddrs[i].af = addrs[i].af;
         laddrs->laddrs[i].addr = addrs[i].addr;


src/ipvs/ip_vs_laddr.c
```diff
@@ -525,6 +525,7 @@ static int get_msg_cb(struct dpvs_msg *msg)
 
     *laddrs = *laddr_conf;
     laddrs->nladdrs = naddr;
+    laddrs->cid = cid;
```
ywc689, line 528: the `laddrs->cid` is set by the user defined input param "conf" in `laddr_sockopt_get`. It's no need to set the param here.

legend050709: the laddrs->cid  set by the user defined input param "conf" is  cpu index , which is not equal to lcore id. 
eg: 
start dpvs with -l  1,2,3,4,5; the cpu index of  slave worker with lcore id 2  is  1.

ywc689: ok.  you are right!

PR #592, opened 2020-06-17
dpip: show dpdk mem info

legend050709:
Currently there is no effective tool to check dpdk memory usage; the commit is used to display dpdk memory usage;
The method of use is as follows：
$dpip eal-mem help
Usage:
    dpip eal-mem show [seg | ring | zone | pool]


legend050709:
@mscbg , @ywc689 
Could you help to review this? Thank you in advance. If you have any advice, please let me know.

PR #593, opened 2020-06-19
conhash: use weight ratio to reduce virtual node num in rbtree

legend050709:
the memory will be a big problem with the conhash scheduling for vs when large count of vs and rs configured per lcore.

### DIFF
--- a/src/ipvs/ip_vs_conhash.c
+++ b/src/ipvs/ip_vs_conhash.c
@@ -25,7 +25,7 @@
 
 struct conhash_node {
     struct list_head    list;
-    uint16_t            weight;
+    uint16_t            weight_ratio;
     int                 af;         /* address family */
     union inet_addr     addr;       /* IP address of the server */
     uint16_t            port;       /* port number of the server */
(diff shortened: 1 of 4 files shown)


src/ipvs/ip_vs_conhash.c
```diff
@@ -182,17 +180,13 @@ static int dp_vs_conhash_add_dest(struct dp_vs_service *svc,
...
-    p_conhash_node->af = dest->af;
-    p_conhash_node->addr = dest->addr;
-    p_conhash_node->port = dest->port;
-    p_conhash_node->weight = weight;
 
     // add node to conhash
     p_node = &(p_conhash_node->node);
-    addr_fold = inet_addr_fold(dest->af, &dest->addr);
-    snprintf(str, sizeof(str), "%u%d", addr_fold, dest->port);
+    inet_ntop(dest->af, &dest->addr, addr, sizeof(addr));
```
legend050709, line 186: for nat66, inet_addr_fold may cause two rs to have the same addr_fold.In this case, there will be an rs that can never be added to the rbtree;

legend050709:
@mscbg , @ywc689 ;  could you help to review this? Thank you in advance. If you have any advice, please let me know.

PR #594, opened 2020-06-21
fix refcount issue when adding/removing VIP group

royhunter:
1. Increase VS's refcount when adding one RS
2. Decrease VS's refcount when deleteing one RS
3. Remove VS when its refcount is zero(No RS associated)


azura27:
@royhunter   the commit "@royhunter
fix refcount issue when adding/removing VIP group" is test OK for the solution of https://github.com/iqiyi/dpvs/issues/631, Thanks a lot  
@ywc689

azura27:
@royhunter  When you are free , please delete the commit of "enable coredump generation" on the PR, for the commit has been PR ed at https://github.com/iqiyi/dpvs/pull/619

liweitianux:
Yes. I've tested that this PR (the refcount commit within it) fixes the keepalived issue #631

PR #595, opened 2020-06-24
Performance issue caused by invoking 'rte_malloc' frequently.

flyinskyabc:
1. Deprecate 'rte_malloc', use 'rte_mempool_get' for inserting UOA instead.
2. Remove unnecessary reference counting for 'struct inet_device'.

### DIFF
--- a/src/ipvs/ip_vs_proto_udp.c
+++ b/src/ipvs/ip_vs_proto_udp.c
@@ -52,6 +52,9 @@ struct conn_uoa {
     uint8_t         acked;
 };
 
+static struct rte_mempool *g_uoa_cache[DPVS_MAX_SOCKET];
+#define this_uoa_cache (g_uoa_cache[rte_socket_id()])
+
 static int g_uoa_max_trail = UOA_DEF_MAX_TRAIL; /* zero to disable UOA */
 static int g_uoa_mode = UOA_M_OPP; /* by default */
 
@@ -148,6 +151,7 @@ static int udp_conn_sched(struct dp_vs_proto *proto,
                         struct dp_vs_conn **conn,
                         int *verdict)
 {
+    int err;
     struct udp_hdr *uh, _udph;
     struct dp_vs_service *svc;
     bool outwall = false;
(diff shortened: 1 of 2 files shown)


src/ipvs/ip_vs_proto_udp.c
```diff
@@ -785,9 +790,60 @@ static int udp_snat_out_handler(struct dp_vs_proto *proto,
...
+        snprintf(pool_name, sizeof(pool_name), "dp_vs_uoa_%d", i);
+
+        g_uoa_cache[i] = rte_mempool_create(pool_name,
+                                            dp_vs_conn_pool_size(),
+                                            sizeof(struct conn_uoa),
+                                            dp_vs_conn_pool_cache_size(),
+                                            0, NULL, NULL, NULL, NULL, i, 0);
+
+        if (unlikely(!g_uoa_cache[i])) {
+            return EDPVS_NOMEM;
```
ywc689, line 808: free previous uoa_cache memory before return

ywc689 (approved):
Codes are OK now. But mempool of the same size with `dp_vs_conn` may not memory friendly enough.

Actually, we are planning to fix rte_malloc problem in all critical packet pathes using dpvs's mempool， which supports variable length mempool and provides a balance between rte_malloc and dpdk mempool.You could help us with the plan if you are interested with it.

src/ipvs/ip_vs_proto_udp.c
```diff
@@ -177,8 +181,8 @@ static int udp_conn_sched(struct dp_vs_proto *proto,
     if ((*conn)->dest->fwdmode == DPVS_FWD_MODE_FNAT && g_uoa_max_trail > 0) {
         struct conn_uoa *uoa;
 
-        (*conn)->prot_data = rte_zmalloc(NULL, sizeof(struct conn_uoa), 0);
-        if (!(*conn)->prot_data) {
+        err = rte_mempool_get(this_uoa_cache, (void **)&((*conn)->prot_data));
```
donghaobo, line 184: for now, struct conn_uoa has 6 bytes only, maybe use prot_data directly
uoa = (void*)&(*conn)->prot_data

PR #600, opened 2020-06-30
add dpdk-pdump filter by five tuple for dpvs.

weiyanhua100:
1.Filter based on the 5-tuple of the packet.
2.Use example: ./dpdk-pdump -- --pdump 'port=0,queue=*,rx-dev=/var/rx.pcap,tx-dev=/var/rx.pcap,proto-port=58818,total-num-mbufs=1025'
3.Open pdump under 20G traffic without packet loss.


azura27:
@weiyanhua100 
hello， when reproducing this patch for DPVS support, the compiling of DPVS does NOT work, if there is some dependencies needing
-----------------------------------------------
(.text+0x161): multiple definition of `__morestack_make_guard'
/usr/lib/gcc/x86_64-redhat-linux/4.8.5/libgcc.a(morestack.o):(.text+0x161): first defined here
collect2: error: ld returned 1 exit status
-----------------------------------------------
would you mind to confirm and provide the workflow and env when you are free, Thanks a lot~

src/config.mk
```diff
@@ -39,12 +39,15 @@ CFLAGS += -D DPVS_MAX_LCORE=64
...
 #CFLAGS += -D CONFIG_DPVS_IPSET_DEBUG
 #CFLAGS += -D CONFIG_NDISC_DEBUG
 #CFLAGS += -D CONFIG_MSG_DEBUG
 #CFLAGS += -D CONFIG_DPVS_MP_DEBUG
 
+ifeq ($(CONFIG_PDUMP), y)
+CFLAGS += -D CONFIG_DPVS_PDUMP
+endif
+
 GCC_MAJOR = $(shell echo __GNUC__ | $(CC) -E -x c - | tail -n 1)
```
ywc689, line 51: What's the purpose to add the`CONFIG_PDUMP`  while `CONFIG_DPVS_PDUMP`  exists.

ywc689:
Accept and merge it. I'll handle the compile problem later.

PR #602, opened 2020-07-05
Send RST when check syn cookie fail.

haosdent:
In LVS, it would send RST when receive invalid packet.

### DIFF
--- a/src/ipvs/ip_vs_synproxy.c
+++ b/src/ipvs/ip_vs_synproxy.c
@@ -957,6 +957,10 @@ int dp_vs_synproxy_ack_rcv(int af, struct rte_mbuf *mbuf,
     struct dp_vs_synproxy_opt opt;
     struct dp_vs_service *svc;
     int res_cookie_check;
+    struct dp_vs_synproxy_opt tcp_opt;
+    struct ether_hdr *eth;
+    struct ether_addr ethaddr;
+    struct netif_port *dev;
 
     /* Do not check svc syn-proxy flag, as it may be changed after syn-proxy step 1. */
     if (!th->syn && th->ack && !th->rst && !th->fin &&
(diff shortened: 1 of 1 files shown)


haosdent:
@mscbg @ywc689 could you help to review this? Thank you in advance.

src/ipvs/ip_vs_synproxy.c
```diff
@@ -986,6 +990,61 @@ int dp_vs_synproxy_ack_rcv(int af, struct rte_mbuf *mbuf,
...
+                RTE_LOG(ERR, IPVS, "%s: netif_xmit failed -- %s\n",
+                        __func__, dpvs_strerror(res));
+                /* should not set verdict to INET_DROP since netif_xmit
+                 * always consume the mbuf while INET_DROP means mbuf'll
+                 * be free in INET_HOOK.*/
+            }
+
+            // NOTE: By right we should return `INET_STOLEN` here but we haven't
+            // find a good way to free `mbuf` if return `INET_STOLEN` directly.
             *verdict = INET_DROP;
```
haosdent, line 1048: Need to change to INET_STOLEN

ywc689: 1. Why not verdict `STOLEN`?  `netif_xmit` would always be consumed the packet. The packet is freed by PMD after xmit. 
2. Wrap the code episodes into a function to keep the codes tidy. For example,
```
int syn_proxy_cookie_fail_send_rst(...)
{
      ...
}
```

ywc689:
The feature is accepted in pr #706.

legend050709:
> In LVS, it would send RST when receive invalid packet.

1）If the syn cookie check fails, just discard the data packet. Why send rst ？
2）If there is an ack attack, will sending a large amount of rst affect performance?

Onepamopa:
I don't know why they haven't made 1) configurable, most people will obviously prefer to discard the packet.
As for 2) - Most likely yes.

legend050709:
> In LVS, it would send RST when receive invalid packet.



> @mscbg @ywc689 could you help to review this? Thank you in advance.



> I don't know why they haven't made 1) configurable, most people will obviously prefer to discard the packet.
> As for 2) - Most likely yes.

@haosdent Can you talk about  the scenario?

haosdent:
Hi, @legend050709 this problem is resolved in https://github.com/iqiyi/dpvs/pull/706

> 1）If the syn cookie check fails, just discard the data packet. Why send rst ？

If we use LVS, kernel would send RST. After we migrate to DPVS, a lot of client hung forever due to this behaviour changes.
Especially when ECMP rehash, previously kernel would send RST and client would re-establish connections immediately, so the impact is small.
But after using DPVS, because DPVS would not send RST, so client hung forever until restart all clients.

> 2）If there is ack attack, will sending a large amount of rst packets affect performance?

Would not be a big issue according to our performance test. But I forgot the exact number, it looks like no differences in our test result.

Onepamopa:
> 
> 
> Hi, @legend050709 this problem is resolved in #706
> 
> > 1）If the syn cookie check fails, just discard the data packet. Why send rst ？
> 
> If we use LVS, kernel would send RST. After we migrate to DPVS, a lot of client hung forever due to this behaviour changes.
> Especially when ECMP rehash, previously kernel would send RST and client would re-establish connections immediately, so the impact is small.
> But after using DPVS, because DPVS would not send RST, so client hung forever until restart all clients.
> 
> > 2）If there is ack attack, will sending a large amount of rst packets affect performance?
> 
> Would not be a big issue according to our performance test. But I forgot the exact number, it looks like no differences in our test result.

Do your tests include actual attacks? Because from what I've seen sending unnecessary packets do impede performance. Our implementation of SYN proxy firewall on top of DPDK has no problems when RST is not sent on invalid SYN-ACK responses. Perhaps there's something wrong in yours?

haosdent:
> SYN proxy firewall 

this is a separate component? I think most people just deploy DPVS.

legend050709:
> > Hi, @legend050709 this problem is resolved in #706
> > > 1）If the syn cookie check fails, just discard the data packet. Why send rst ？
> > 
> > 
> > If we use LVS, kernel would send RST. After we migrate to DPVS, a lot of client hung forever due to this behaviour changes.
> > Especially when ECMP rehash, previously kernel would send RST and client would re-establish connections immediately, so the impact is small.
> > But after using DPVS, because DPVS would not send RST, so client hung forever until restart all clients.
> > > 2）If there is ack attack, will sending a large amount of rst packets affect performance?
> > 
> > 
> > Would not be a big issue according to our performance test. But I forgot the exact number, it looks like no differences in our test result.
> 
> Do your tests include actual attacks? Because from what I've seen sending unnecessary packets does impede performance. Our implementation of SYN proxy firewall on top of DPDK has no problems when RST is not sent on invalid ACK responses. Perhaps there's something wrong in yours?

I have two questions:
(1）the ack packet in three-way handshake  sent by a normal client will not fail the syn cookie check.
Only the fake ack package will fail to check, so why sending rst to a  fake client at this time?
(2)  “If we use LVS, kernel would send RST.” Which kernel version are you talking about?
I do not find the operation of sending rst when syn cookied checking failed by synproxy_recv_client_ack-->__cookie_v4_check of kernel 4.18.0

haosdent:
hi, @legend050709 you could try `hping3 -A -p 80 ` in a normal LVS server and DPVS server to compare the differences.

PR #608, opened 2020-07-20
Fix icmp outbound address family bug.

SimonCZW:
In the nat64 scenario, backend may reply ICMP_TIME_EXCEEDED packet.  ICMP_TIME_EXCEEDED packet's call chain：`__dp_vs_in`->`dp_vs_in_icmp`->`__dp_vs_in_icmp4`->`xmit_outbound_icmp`, in `xmit_outbound_icmp` will invoke `__xmit_outbound_icmp6` by af from `conn->af` (which is ipv6 in nat64 scenario), and it will invoke ipv6 function for ipv4 packet.

Same reason change for `dp_vs_xmit_icmp` function.

### DIFF
--- a/src/ipvs/ip_vs_core.c
+++ b/src/ipvs/ip_vs_core.c
@@ -515,7 +515,10 @@ static int xmit_outbound_icmp(struct rte_mbuf *mbuf,
                               struct dp_vs_proto *prot,
                               struct dp_vs_conn *conn)
 {
-    int af = conn->af;
+    struct conn_tuple_hash *t;
+
+    t = &tuplehash_out(conn);
+    int af = t->af;
 
     assert(af == AF_INET || af == AF_INET6);
     if (af == AF_INET)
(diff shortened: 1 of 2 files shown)


ywc689:
I do not think the problem is solved completely.
1. mbuf should also be transformed for nat64
2. inbound icmp traffic also has problem

SimonCZW:
> 1. mbuf should also be transformed for nat64

Solution: nat46 for inner packet and wrapped by icmpv6 header(ICMP_TIME_EXCEEDED). Am i get your point? 

> 2. inbound icmp traffic also has problem

DPVS do not support nat46 for snat. inbound icmp traffic should be ok?

ywc689:
In case of nat64, ICMP_TIME_EXCEEDED packet recieved from RS is IPv4. It need to transformed to IPv6 packets when forwarding to Client. Maybe something like "__dp_vs_xmit_icmp64" and "__dp_vs_xmit_icmp46" are needed.

SimonCZW:
Thanks, I will figure out.

SimonCZW:
> In case of nat64, ICMP_TIME_EXCEEDED packet recieved from RS is IPv4. It need to transformed to IPv6 packets when forwarding to Client. Maybe something like "__dp_vs_xmit_icmp64" and "__dp_vs_xmit_icmp46" are needed.

ICMP cannot be nat64 because icmp and icmpv6 is too difference in L4 layer. This bug seem cannot be fix but just avoid error.

PR #612, opened 2020-07-22
dpvs: add the support of 4 numa-node server like below.

andrewhit:
```
> lscpu
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                96
On-line CPU(s) list:   0-47
Off-line CPU(s) list:  48-95
Thread(s) per core:    1
Core(s) per socket:    24
Socket(s):             2
NUMA node(s):          4
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 85
Model name:            Intel(R) Xeon(R) Platinum 8260 CPU @ 2.40GHz
Stepping:              7
CPU MHz:               3099.999
CPU max MHz:           3900.0000
CPU min MHz:           1000.0000
BogoMIPS:              4800.00
Virtualization:        VT-x
L1d cache:             32K
L1i cache:             32K
L2 cache:              1024K
L3 cache:              36608K
NUMA node0 CPU(s):     0-3,6-8,12-14,19,20
NUMA node1 CPU(s):     4,5,9-11,15-18,21-23
NUMA node2 CPU(s):     24-27,31-33,37-39,43,44
NUMA node3 CPU(s):     28-30,34-36,40-42,45-47
...
```


ywc689:
Macros defined in "src/config.mk" need to change according to the environments where you run dpvs. You should configure them before running `make` according to your needs.
Besides, increasing "DPVS_MAX_SOCKET" will increase dpvs memory in multiples. 2 numa-nodes server is used more widely now.  Using 2 as the default value is reasonable.

PR #615, opened 2020-07-23
fix build error in gcc 7.3.0

wangxp006:
Fix build error
/root/dpvs/dpvs-master/src/iftraf.c: In function ‘iftraf_sockopt_get’:
/root/dpvs/dpvs-master/src/iftraf.c:388:5: error: this ‘if’ clause does not guard... [-Werror=misleading-indentation]
     if (!conf || size < sizeof(struct dp_vs_iftraf_conf) || !out || !outsize)
     ^~
/root/dpvs/dpvs-master/src/iftraf.c:390:2: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the ‘if’
  cf = conf;
  ^~

### DIFF
--- a/src/iftraf.c
+++ b/src/iftraf.c
@@ -387,7 +387,7 @@ int iftraf_sockopt_get(sockoptid_t opt, const void *conf, size_t size,
 
     if (!conf || size < sizeof(struct dp_vs_iftraf_conf) || !out || !outsize)
         return EDPVS_INVAL;
-	cf = conf;
+    cf = conf;
 
     if (cf && strlen(cf->ifname)) {
         port = netif_port_get_by_name(cf->ifname);


wangxp006:
[root@localhost dpvs-master]# gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-linux-gnu/7.3.0/lto-wrapper
Target: x86_64-linux-gnu
Configured with: ../configure --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --enable-shared --enable-threads=posix --enable-checking=release --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-linker-hash-style=gnu --enable-languages=c,c++,objc,obj-c++,fortran,lto --enable-plugin --enable-initfini-array --disable-libgcj --without-isl --without-cloog --enable-gnu-indirect-function --build=x86_64-linux-gnu --with-stage1-ldflags=' -Wl,-z,relro,-z,now' --with-boot-ldflags=' -Wl,-z,relro,-z,now' --with-tune=generic --with-arch_32=x86-64 --disable-multilib
Thread model: posix
gcc version 7.3.0 (GCC)

ywc689:
The problem has been fixed in devel branch. We'll merge it into master branch in next release.

PR #617, opened 2020-07-27
One Packet Scheduler for UDP

donghaobo:
from kernel ipvs

### DIFF
--- a/include/ipvs/conn.h
+++ b/include/ipvs/conn.h
@@ -44,6 +44,7 @@ enum {
     DPVS_CONN_F_SYNPROXY         = 0x8000,
     DPVS_CONN_F_TEMPLATE         = 0x1000,
     DPVS_CONN_F_NOFASTXMIT       = 0x2000,
+    DPVS_CONN_F_ONE_PACKET       = 0x4000,
 };
 
 struct dp_vs_conn_param {
--- a/include/ipvs/conn.h
+++ b/include/ipvs/conn.h
@@ -44,7 +44,7 @@ enum {
     DPVS_CONN_F_SYNPROXY         = 0x8000,
     DPVS_CONN_F_TEMPLATE         = 0x1000,
     DPVS_CONN_F_NOFASTXMIT       = 0x2000,
-    DPVS_CONN_F_ONE_PACKET       = 0x4000,
+    DPVS_CONN_F_ONE_PACKET       = 0x0400,
 };
 
 struct dp_vs_conn_param {
--- a/src/ipvs/ip_vs_conn.c
+++ b/src/ipvs/ip_vs_conn.c
@@ -181,8 +181,7 @@ static void dp_vs_conn_refresh_timer(struct dp_vs_conn *conn, bool lock)
         return;
 
     if (conn->flags & DPVS_CONN_F_ONE_PACKET) {
-        conn->timeout.tv_sec = 0;
-        conn->timeout.tv_usec = 0;
+        return;
     }
 
     if (dp_vs_conn_is_template(conn)) {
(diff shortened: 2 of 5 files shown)


ytwang0320:
It seems not complete.
Add this "one-packet scheduling" option in  ipvsadm and keepalived.

donghaobo:
> It seems not complete.
> Add this "one-packet scheduling" option in ipvsadm and keepalived.

ipvsadm and keepalived already support "one-packet scheduling" , config with "ops"

tools/ipvsadm/ipvsadm.c
```diff
@@ -1907,7 +1907,7 @@ print_service_entry(ipvs_service_entry_t *se, unsigned int format, lcoreid_t cid
 		if (se->pe_name[0])
 			printf(" pe %s", se->pe_name);
 		if (se->user.flags & IP_VS_SVC_F_ONEPACKET)
-			printf(" ops");
+			printf(" --ops");
```
donghaobo, line 1910: add '--' to make "ops" loadable

src/ipvs/ip_vs_conn.c
```diff
@@ -177,6 +180,11 @@ static void dp_vs_conn_refresh_timer(struct dp_vs_conn *conn, bool lock)
     if (!dp_vs_conn_is_in_timer(conn))
         return;
 
+    if (conn->flags & DPVS_CONN_F_ONE_PACKET) {
+        conn->timeout.tv_sec = 0;
+        conn->timeout.tv_usec = 0;
+    }
```
ywc689, line 186: One-packet-conn does not need timer . Why not return here ?

src/ipvs/ip_vs_conn.c
```diff
@@ -1150,6 +1166,12 @@ void dp_vs_conn_put_no_reset(struct dp_vs_conn *conn)
 /* put back the conn and reset it's timer */
 void dp_vs_conn_put(struct dp_vs_conn *conn)
 {
+    if ((conn->flags & DPVS_CONN_F_ONE_PACKET) &&
+            (rte_atomic32_read(&conn->refcnt) == 1) &&
+            !dp_vs_conn_is_in_timer(conn)) {
```
ywc689, line 1171: I think `(rte_atomic32_read(&conn->refcnt) == 1)` and `!dp_vs_conn_is_in_timer(conn)` are not needed here.

include/ipvs/conn.h
```diff
@@ -44,6 +44,7 @@ enum {
     DPVS_CONN_F_SYNPROXY         = 0x8000,
     DPVS_CONN_F_TEMPLATE         = 0x1000,
     DPVS_CONN_F_NOFASTXMIT       = 0x2000,
+    DPVS_CONN_F_ONE_PACKET       = 0x4000,
```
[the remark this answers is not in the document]
donghaobo: actually `DPVS_CONN_F_EXPIRE_QUIESCENT`  should be `DP_VS_SVC_F_EXPIRE_QUIESCENT`, and it is used in ip_vs_core.c   `conn->dest->svc->flags & DPVS_CONN_F_EXPIRE_QUIESCENT`

donghaobo:
> What are the use case for one packet schedule? Some particular processes are added to DPVS critical packet forwarding path. This may harm the performance. We should make a tradeoff between functionality and performance.

the client may generate too much syslogs (or any other logs), which cannot be handled by one server.

src/ipvs/ip_vs_conn.c
ywc689: 1. The conditions ` (rte_atomic32_read(&conn->refcnt) == 1) &&  !dp_vs_conn_is_in_timer(conn)` should always be true. You can verify it with `assert`. If no problem after test, it's advised to remove the unecessary conditions here because `dp_vs_conn_put` impacts performance a lot.
2. `dp_vs_conn_put_nolock` should modified accordingly.

ywc689 (approved):

PR #619, opened 2020-07-29
enable coredump generation

royhunter:
Enable coredump generation via rlimit setting from code level

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -19,6 +19,7 @@
 #include <pthread.h>
 #include <assert.h>
 #include <getopt.h>
+#include <sys/resource.h>
 #include "pidfile.h"
 #include "dpdk.h"
 #include "conf/common.h"
@@ -139,6 +140,14 @@ static int parse_app_args(int argc, char **argv)
     return ret;
 }
 
+static void enable_coredump(void)
+{
+    struct rlimit core_limits;
+
+    core_limits.rlim_cur = core_limits.rlim_max = RLIM_INFINITY;
+    setrlimit(RLIMIT_CORE, &core_limits);
+}
+
 int main(int argc, char *argv[])
 {
     int err, nports;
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -148,6 +157,8 @@ int main(int argc, char *argv[])
     char pql_conf_buf[LCORE_CONF_BUFFER_LEN];
     int pql_conf_buf_len = LCORE_CONF_BUFFER_LEN;
 
+    enable_coredump();
+
```
ywc689, line 161: It's advised to make some improvements for this feature.
1. Coredump feature needs to be enabled/disabled from user config files or command option.
2. Coredump soft/hard resource limits should be set by users.

andrewhit:
> It's advised to make some improvements for this feature.
> 
> 1. Coredump feature needs to be enabled/disabled from user config files or command option.
> 2. Coredump soft/hard resource limits should be set by users.

Even though the below configuration is added, it does not guarantee the coredump is determined to be generated. So this patch is provided.

```
root@n25-146-090:/# cat /etc/security/limits.conf
# /etc/security/limits.conf
...
root    soft    core    unlimited
```

```
root@n25-146-090:/# cat /etc/rc.local
#!/bin/sh -e
modprobe ipmi_devintf; modprobe ipmi_si
echo "/opt/tiger/cores/core.%e.%p.%h.%t" > /proc/sys/kernel/core_pattern
ulimit -c unlimited
exit 0
```

ywc689:
Coredump configured in "/etc/security/limits.conf" may be ignored if your DPVS is managed by a service manager, such as `systemd`. In this case, you should enable coredump by the service manager rather than the limits.conf file. Refer to https://zhuanlan.zhihu.com/p/41153588 as an example.

ywc689:
Coredump can be enabled/disabled by the means outside of DPVS.  This patch is closing, and if any problem, please reopen it.

PR #634, opened 2020-08-12
add proxy_protocol_v2 support for fnat.

helloliupan:
'proxy_protocol' keyword should be configured in 'keepalived.conf', and virtual_server should be configured with FNAT for TCP, the detailed configurations are as follows:
1. 'global_defs': proxy_protocol_v2 support is available for all virtual_server when 'proxy_protocol' is configured in 'global_defs' block:
global_defs {
        ......
        proxy_protocol
        ......
}

2. 'virtual_server': proxy_protocol_v2 support is available for the given virtual_server when 'proxy_protocol' is configured in 'virtual_server' block:
virtual_server ip_addr port {
        ........
        lb_kind FNAT
        protocol TCP
        proxy_protocol
        ........
}

### DIFF
--- a/include/conf/pphdr.h
+++ b/include/conf/pphdr.h
@@ -1,0 +1,66 @@
+/*
+ * DPVS is a software load balancer (Virtual Server) based on DPDK.
+ *
+ * Copyright (C) 2017 iQIYI (www.iqiyi.com).
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ */
+#ifndef __DPVS_PPHDR_H__
+#define __DPVS_PPHDR_H__
+
+#include <netinet/tcp.h>
+#include <stdint.h>
+
+#define PROXY_PROTO_HDR_LEN_V4 28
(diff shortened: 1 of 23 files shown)


ywc689:
Could you explain the use cases and the implementation idea in brief?   I think TOA can be used for the same purpose.

ywc689:
DPVS delivers source IP/port  to backends by TOA. Proxy protocol do almost the same thing in another way. It's suggested that this feature should be enabled by a macro (such as "CONFIG_PROXY_PROTOCOL_V2") defined in "src/config.mk" so that it wouldn't cause overhead for  the one who don't need this feature.

donghaobo:
seems that PP retrans is not considered

helloliupan:
> DPVS delivers source IP/port to backends by TOA. Proxy protocol do almost the same thing in another way. It's suggested that this feature should be enabled by a macro (such as "CONFIG_PROXY_PROTOCOL_V2") defined in "src/config.mk" so that it wouldn't cause overhead for the one who don't need this feature.

thanks for you suggestion. Proxy protocol(PP) has some advantageses over TOA as follow:
1. PP can deliver source IP/Port and dest IP/Port to real servers with both IPv4 and IPv6 condition, while TOA can not fill both source and dest addrs with IPv6 situation;
2. PP is an application layer protocol, we do not need to adapt different kernel versions if the real servers support PP at the same while, nginx for example(https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol/).

helloliupan:
> seems that PP retrans is not considered

Right, the retrans for first packet of application layer from client may pass to real servers without PP header, I will fix this later.

donghaobo:
> Right, the retrans for first packet of application layer from client may pass to real servers without PP header, I will fix this later.

if pp header pkt lost, the connection will retry until uplayer close.
client will retrans first payload again and again,  server will ack -PP_LEN again and again.

ywc689:
#923 is the full implementation for proxy protocol in DPVS.

PR #645, opened 2020-08-26
dpvs optimize

yiflv:


src/netif.c
```diff
@@ -2472,16 +2472,31 @@ static void lcore_job_xmit(void *args)
...
+
+static void lcore_job_slave_timer_manage(void *args) 
+{
+    int64_t local_ticks;
+    lcoreid_t cid = rte_lcore_id();    
+
+    local_ticks = rte_atomic64_read(&ticks);
+    while (local_ticks != ticks_tbl[cid]) {
+        rte_timer_manage();
+        ticks_tbl[cid]++;
```
ywc689, line 2486: `rte_timer_manage` may be called more than once for each loop. If so, it's not fair for other jobs.

ywc689:
Timer precision of worker lcore may be affect by master lcore. Test and verify the impact.

PR #648, opened 2020-08-28
modify alive juding when alive_p overflow

azura27:
for the issue 631, the alive_p would be overflow after "--" when alive_p equal 0
So I modify the VS alive jugement, I do not find other method to solve this issue uptodate.

---------background  before chaging--------------------
using quorum script with VIP param delivering would not cause this issue to happen
not using quorum would cause this issue when change VIP addr alone for two times or more

### DIFF
--- a/tools/keepalived/keepalived/check/ipvswrapper.c
+++ b/tools/keepalived/keepalived/check/ipvswrapper.c
@@ -300,19 +300,18 @@ ipvs_group_range_cmd(int cmd, ipvs_service_t *srule, ipvs_dest_t *drule, virtual
 static bool
 is_vsge_alive(virtual_server_group_entry_t *vsge, virtual_server_t *vs)
 {
-	unsigned t = -1;
 	if (vsge->is_fwmark) {
 		if (vs->af == AF_INET)
-			return (vsge->fwm4_alive == t ? 0 : !!vsge->fwm4_alive);
+			return !!vsge->fwm4_alive;
 		else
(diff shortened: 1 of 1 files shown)


tools/keepalived/keepalived/check/ipvswrapper.c
```diff
@@ -300,18 +300,19 @@ ipvs_group_range_cmd(int cmd, ipvs_service_t *srule, ipvs_dest_t *drule, virtual
 static bool
 is_vsge_alive(virtual_server_group_entry_t *vsge, virtual_server_t *vs)
 {
+	unsigned t = -1;
```
ywc689, line 303: It's a little strange to assign -1 to an unsigned variable.

ywc689 (commented):
Can you figure out the root cause of the underflow problem? I think logical problem exists somewhere and we'd better find it out.

tools/keepalived/keepalived/check/ipvswrapper.c
azura27: yes，do you have any suggestion on overflow judgement

ytwang0320:
Please delete `unset_vsge_alive` in func `ipvs_group_remove_entry` and have a test.

azura27:
> Please delete `unset_vsge_alive` in func `ipvs_group_remove_entry` and have a test.

Test again, this worked for "virtual_server_group" with single virtual_server instance
But for multi  virtual_server instances under  "virtual_server_group", the virtual_server been left again

PR #649, opened 2020-08-28
ipvsadm: fix change conhash target error

legend050709:
If the hash factor of the conhash scheduling algorithm of vs is sip, we want to change the hash factor to qid through ipvsadm -E -u xxx -s conhash -Y qid. we will find that the vs will have two hash factors, sip and qid through ipvsadm -ln.

### DIFF
--- a/tools/keepalived/keepalived/check/libipvs.c
+++ b/tools/keepalived/keepalived/check/libipvs.c
@@ -243,6 +243,8 @@ int ipvs_update_service_by_options(ipvs_service_t *svc, unsigned int options)
 	}
 
 	if (options & OPT_HASHTAG) {
+		app.user.flags &= ~ IP_VS_SVC_F_QID_HASH;
+		app.user.flags &= ~ IP_VS_SVC_F_SIP_HASH;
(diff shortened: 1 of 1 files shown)


tools/keepalived/keepalived/check/libipvs.c
```diff
@@ -243,6 +243,8 @@ int ipvs_update_service_by_options(ipvs_service_t *svc, unsigned int options)
 	}
 
 	if (options & OPT_HASHTAG) {
+		app.user.flags &= ~ IP_VS_SVC_F_QID_HASH;
+		app.user.flags &= ~ IP_VS_SVC_F_SIP_HASH;
```
ywc689, line 247: Line 209~210 cleared these two option flags. Was it not working?

legend050709: > Line 209~210 cleared these two option flags. Was it not working?

The purpose on 209 and 210 is to clear the sip and qid flag  configuration obtained from dpvs when the scheduling algorithm is not conhash

ywc689 (approved):

PR #655, opened 2020-09-02
snat match use acl

donghaobo:
snat match use acl

### DIFF
--- a/src/ipvs/ip_vs_service.c
+++ b/src/ipvs/ip_vs_service.c
@@ -52,31 +52,8 @@ static struct list_head dp_vs_svc_table[DPVS_MAX_LCORE][DP_VS_SVC_TAB_SIZE];
 
 static struct list_head dp_vs_svc_fwm_table[DPVS_MAX_LCORE][DP_VS_SVC_TAB_SIZE];
 
-/* lock for master lcore */
-static rte_rwlock_t dp_vs_svc_match_lock[DPVS_MAX_LCORE];
 static struct hlist_head dp_vs_svc_match_list[DPVS_MAX_LCORE][DP_VS_SVC_TAB_SIZE];
 
-/* this is for match acl thread */
-int dp_vs_services_match_iter(int (*cb)(struct dp_vs_service *, void*),
-                     void* data, lcoreid_t cid)
-{
-    int idx = 0;
-    struct dp_vs_service *svc = NULL;
-    int ret = EDPVS_OK;
-
-    rte_rwlock_read_lock(&dp_vs_svc_match_lock[cid]);
-    for (idx = 0; idx < DP_VS_SVC_TAB_SIZE; idx++) {
-        hlist_for_each_entry(svc, &dp_vs_svc_match_list[cid][idx], m_list) {
-            ret = cb(svc, data);
-            if (ret != EDPVS_OK) {
-                rte_rwlock_read_unlock(&dp_vs_svc_match_lock[cid]);
-                return ret;
-            }
-        }
-    }
-    rte_rwlock_read_unlock(&dp_vs_svc_match_lock[cid]);
-    return EDPVS_OK;
-}
 
 static inline int dp_vs_service_hashkey(int af, unsigned proto, const union inet_addr *addr)
 {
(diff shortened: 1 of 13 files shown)


src/ipvs/ip_vs_service_match_acl.c
```diff
@@ -0,0 +1,929 @@
...
+        dp_vs_svc_match_acl4[cid] = NULL;
+        dp_vs_svc_match_acl6[cid] = NULL;
+    }
+    stop_build = 0;
+    pthread_t tid;
+    if (pthread_create(&tid, NULL, dpvs_svc_match_build_loop, NULL)) {
+        RTE_LOG(ERR, IPVS, "fail to create dpvs_svc_match_build_loop thread\n");
+        dp_vs_svc_match_term();
+        return EDPVS_SYSCALL;
+    }
```
ywc689, line 877: DPVS is based on a poll mode driver, with each thread pinned to an exclusive cpu, and extra threading is not encouraged for performance considerations.

src/ipvs/ip_vs_service.c
```diff
@@ -36,21 +37,46 @@
...
+#include "ipvs/service_match_acl.h"
 
-static rte_atomic16_t dp_vs_num_services[DPVS_MAX_LCORE];
+static rte_atomic32_t dp_vs_num_services[DPVS_MAX_LCORE];
 
 /**
  * hash table for svc
  */
-#define DP_VS_SVC_TAB_BITS 8
+#define DP_VS_SVC_TAB_BITS 14
```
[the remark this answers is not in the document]
donghaobo: we are trying to add 100k+ snat svc, 8-bit is not enough

src/ipvs/ip_vs_service.c
```diff
@@ -36,21 +37,46 @@
...
 #define DP_VS_SVC_TAB_SIZE (1 << DP_VS_SVC_TAB_BITS)
 #define DP_VS_SVC_TAB_MASK (DP_VS_SVC_TAB_SIZE - 1)
 
 static struct list_head dp_vs_svc_table[DPVS_MAX_LCORE][DP_VS_SVC_TAB_SIZE];
 
 static struct list_head dp_vs_svc_fwm_table[DPVS_MAX_LCORE][DP_VS_SVC_TAB_SIZE];
 
-static struct list_head dp_vs_svc_match_list[DPVS_MAX_LCORE];
+/* lock for master lcore */
+static rte_rwlock_t dp_vs_svc_match_lock[DPVS_MAX_LCORE];
```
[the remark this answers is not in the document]
donghaobo: this lock is for svc match ACL build thread.  and it is per-core, will be used only at svc update. dataplan will not use it. 
i know the svc operation is processed in slave workers, this lock will slow down the svc insert/delete a little. 
but this lock is  hard to remove. we are planing to modify the control plane with RCU, and that is another project.

src/ipvs/ip_vs_service_match_acl.c
donghaobo: 10w dpdk ACL rules compiling will take about 10 seconds, this should not do in the master. a new thread is very needed.

SimonCZW:
How much performance has been improved with this solution? @donghaobo

donghaobo:
> How much performance has been improved with this solution? @donghaobo

this commit is used with 10k+ SNAT rules。it's trying to reach the same performance as only one SNAT rule . if there's only a default SNAT rule, this commit does nothing help.
if you really run dpvs with 10k+ SNAT rules, you will know how bad the performance is .

src/netif.c
```diff
@@ -1120,7 +1120,8 @@ int dpvs_wait_lcores(void)
     do {
         sched_yield();
         for (i = 0; i < DPVS_MAX_LCORE; i++) {
-            if (is_lcore_id_fwd(i) && loop[i] == lcore_stats[i].lcore_loop) {
```
donghaobo: pr #669  need this change

yifan1227:
May I ask in what circumstances 10k+ SNAT rules are needed?  How many public IPs do you configure for each SNAT director?

PR #657, opened 2020-09-04
show dpvs's ip in traceroute path

donghaobo:
when `fl4->fl4_saddr.s_addr == htonl(INADDR_ANY) ` ,source ip should be selected after route lookup.


ywc689:
It seems this pr trying to solve the same problem as Pr #630 . You can review and test whether it works.

donghaobo:
> It seems this pr trying to solve the same problem as Pr #630 . You can review and test whether it works.

yes, same problem, but different bugs

ywc689 (approved):

PR #660, opened 2020-09-15
fix the dest max conn bug

legend050709:
Two questions:
1) the max_conn of  dest should be based on the whole machine, not based on the lcore;
2) Changing the uthreshold of rs in keepalived.conf does not take effect;

### DIFF
--- a/src/ipvs/ip_vs_dest.c
+++ b/src/ipvs/ip_vs_dest.c
@@ -52,19 +52,28 @@ static void __dp_vs_dest_update(struct dp_vs_service *svc,
                                 struct dp_vs_dest_conf *udest)
 {
     int conn_flags;
+    uint8_t num_lcores;
 
+    netif_get_slave_lcores(&num_lcores, NULL);
     rte_atomic16_set(&dest->weight, udest->weight);
     conn_flags = udest->conn_flags | DPVS_CONN_F_INACTIVE;
-
     rte_atomic16_set(&dest->conn_flags, conn_flags);
-
 
     dp_vs_dest_set_avail(dest);
 
     if (udest->max_conn == 0 || udest->max_conn > dest->max_conn)
         dest->flags &= ~DPVS_DEST_F_OVERLOAD;
-    dest->max_conn = udest->max_conn;
-    dest->min_conn = udest->min_conn;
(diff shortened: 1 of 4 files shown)


src/ipvs/ip_vs_dest.c
```diff
@@ -63,8 +74,12 @@ static void __dp_vs_dest_update(struct dp_vs_service *svc,
...
     if (udest->max_conn == 0 || udest->max_conn > dest->max_conn)
         dest->flags &= ~DPVS_DEST_F_OVERLOAD;
-    dest->max_conn = udest->max_conn;
-    dest->min_conn = udest->min_conn;
+    dest->max_conn = udest->max_conn / num_lcores;
+    dest->min_conn = udest->min_conn / num_lcores;
+    if (rte_lcore_id() == first_lcore_cid) {
+        dest->max_conn += udest->max_conn % num_lcores;
+        dest->min_conn += udest->min_conn % num_lcores;
+    }
```
ywc689, line 82: I think exact max/min_conn don't make sense because the conns on different workers are just coarsely the same.
Considering the worker number is much smaller the conns, maybe something like the following is enough.

```
dest->max_conn = (udest->max_conn+num_lcores) / num_lcores;
dest->min_conn = (udest->min_conn+num_lcores)  / num_lcores;
```

legend050709: Yes, max_conn is generally much larger than the number of lcore. 
If we use dest->max_conn = (udest->max_conn+num_lcores -1) / num_lcores;
the problem is that the max_conn displayed through ipvsadm -ln --thresholds is not very friendly, 
because the displayed max_conn is not consistent with the actual configured max_conn.

ywc689: We don't set conn limits in most cases, where the `udest->max_conn` and `udest->min_conn` are zero.
In consideration of performance,  we can spare the efforts of calculating per-lcore's conn limit values when conn limits are not specified.

legend050709: I have removed the loop for getting the first forwarding lcore, put the excess conn to the rs in the master lcore, to ensure that the sum of max conn is consistent with the configured max conn.

ywc689 (approved):

PR #662, opened 2020-09-18
kmod/toa: fix is_ro_addr()

yongjianchn:
#484 by @goecho was trying to fix crash problem on some kernel versions. The new function is_ro_addr() is used to check the `address` is `ro` or `rw`.

but the condition seems weired:
```
if (pte->pte &~ _PAGE_RW) {
  ro_enable = 1;
}
```

`pte->pte &~ _PAGE_RW` means there are some other flags being set, no matter RW flags is set or not.

### DIFF
--- a/kmod/toa/toa.c
+++ b/kmod/toa/toa.c
@@ -96,11 +96,11 @@ struct toa_stats_entry toa_stats[] = {
 unsigned int is_ro_addr(unsigned long addr)
 {
     unsigned int level;
-    unsigned int ro_enable = 0;
+    unsigned int ro_enable = 1;
     pte_t *pte = lookup_address(addr, &level);
-    if (pte->pte &~ _PAGE_RW)
+    if (pte->pte & _PAGE_RW)
     {
-            ro_enable = 1;
+            ro_enable = 0;
     }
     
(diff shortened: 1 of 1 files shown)


ywc689:
It seems good to me. But I wonder why a lot of others also do as this? Did you test your patch? Did it work OK?

https://stackoverflow.com/questions/2103315/linux-kernel-system-call-hooking-example
https://lwn.net/Articles/584198/

yongjianchn:
With #484 , I got crash on linux-4.9 (build by myself).
Then I found the wrong `if` condition, which "thinks" the related addrs is RO, and it did the `set_rw()` and `set_ro()`，then seconds later, my kernel crashed.
Actually, the addrs is `RW`, I changed the `if` condition (like this pull request), everything works fine.

I also test the latest CentOS 7, which also works fine.

yongjianchn:
I changed base branch from master to devel.

Test on:

* the latest CentOS 7 (kernel-3.10.0-1127.el7.x86_64)
* the latest CentOS 8 (kernel-4.18.0-193.19.1.el8_2.x86_64)

ywc689 (approved):

PR #667, opened 2020-10-23
Use the same RTE_MACHINE and CPU CFLAGS as DPDK

liweitianux:
DPVS was hardcoding `RTE_MACHINE=native` and CPU CFLAGS in `src/dpdk.mk`.
However, these can be different from the values chosen by DPDK, which
may be a potential issue.  Especially, the `RTE_MACHINE=native` CFLAG
is preventing a portable build.  For example, I couldn't build DPVS on
one machine and deploy it on another machine when their CPU generation
is different (DPVS is unable to run due to `illegal instruction` error).

So, it's better to include `$(RTE_SDK)/mk/rte.vars.mk` and then DPVS
would be built with the same `RTE_MACHINE` and CPU CFLAGS as DPDK.  This
is also used by DPDK examples.

In addition, use `$(Q)` (provided by `rte.vars.mk`) instead of `@` in
`src/Makefile`, which allows us to pass `V=1` to make to show verbose
build commands, similar to the DPDK build.


ywc689 (approved):

ywc689:
Compile failed. The error:

```
/home/wencyu/github/dpvs/src/main.c: In function ‘set_all_thread_affinity’:
/home/wencyu/github/dpvs/src/main.c:60:5: error: implicit declaration of function ‘CPU_ZERO’ [-Werror=implicit-function-declaration]
     CPU_ZERO(&cpuset);
     ^
/home/wencyu/github/dpvs/src/main.c:62:9: error: implicit declaration of function ‘CPU_SET’ [-Werror=implicit-function-declaration]
         CPU_SET(cid, &cpuset);
         ^
/home/wencyu/github/dpvs/src/main.c:64:5: error: implicit declaration of function ‘pthread_setaffinity_np’ [-Werror=implicit-function-declaration]
     s = pthread_setaffinity_np(tid, sizeof(cpu_set_t), &cpuset);
     ^
/home/wencyu/github/dpvs/src/main.c:72:5: error: implicit declaration of function ‘pthread_getaffinity_np’ [-Werror=implicit-function-declaration]
     s = pthread_getaffinity_np(tid, sizeof(cpu_set_t), &cpuset);
     ^
/home/wencyu/github/dpvs/src/main.c:80:9: error: implicit declaration of function ‘CPU_ISSET’ [-Werror=implicit-function-declaration]
         if (CPU_ISSET(cid, &cpuset))
         ^
cc1: all warnings being treated as errors
make[1]: *** [/home/wencyu/github/dpvs/src/main.o] Error 1
make[1]: Leaving directory `/home/wencyu/github/dpvs/src'
make: *** [all] Error 1
```

`_GNU_SOURCE` macro may not be removed.

liweitianux:
Interesing.  I've tested the latest devel DPVS with DPDK 18.11.2 (built with plain make, not meson) and it worked fine.

The `_GNU_SOURCE` has been defined in DPDK's config/mk, and the compiler would report the `_GNU_SOURCE` has been duplicate defined.

What DPDK version are you using?  I'll try to reproduce your build combo and update the patch.

liweitianux:
See [DPDK/v18.11/mk/target/generic/rte.vars.mk](https://github.com/DPDK/dpdk/blob/v18.11/mk/target/generic/rte.vars.mk#L111)

See also DPDK commit: [mk: build with _GNU_SOURCE defined by default](https://github.com/DPDK/dpdk/commit/5d7b673d5fd6663b20c675dd382d9fb43b42af18)

So I'll update my patch to use something like:
```
#ifndef _GNU_SOURCE
#define _GNU_SOURCE
#endif
```
and retry.

Cheers,

ywc689:
Yes. The compile error is from an ealier DPDK version 17.11.2. I've compiled the patch on 18.11.6, everything seems OK.  The conditional macro you post above would fix it.

liweitianux:
Hi @ywc689 , I've updated my patch and it should now be fine.

Cheers

ywc689 (approved):

liweitianux:
Thank  you :smiley:

src/Makefile
```diff
@@ -34,8 +34,8 @@ DATE_STRING := $(shell date +%Y.%m.%d.%H:%M:%S)
 # same path of THIS Makefile
 SRCDIR := $(dir $(realpath $(firstword $(MAKEFILE_LIST))))
 
-include $(SRCDIR)/config.mk
 include $(SRCDIR)/dpdk.mk
+include $(SRCDIR)/config.mk
```
legend050709, line 38: what is the purpose to  move include $(SRCDIR)/config.mk after  $(SRCDIR)/dpdk.mk?
it may cause the problem that CONFIG_MLX5 not work after changing the order.

liweitianux: Well, the `$(RTE_SDK)/mk/rte.vars.mk` included in `dpdk.mk` overrides `CFLAGS`, so we need to include `dpdk.mk` first. Maybe `CONFIG_MLX5` should be handled in another way that works well with the new include ordering.

copybash: 编译说明的文档不打算提一下 CONFIG_MLX5 吗？我用 1.8.8，MLX5_CORE PMD，按照入门文档，搞了 3 天才发现 原来还有 CONFIG_MLX5 这个配置项卡了一道

PR #668, opened 2020-10-27
create a file, mark that the reload is done

donghaobo:
create a file, mark that the reload is done, and inform the reload script. 

reload may be wrong if we don't know when the reload is done and changed the config file for next reload

### DIFF
--- a/tools/keepalived/keepalived/check/check_daemon.c
+++ b/tools/keepalived/keepalived/check/check_daemon.c
@@ -73,6 +73,7 @@
 #include "track_process.h"
 #endif
 
+extern char *reload_done_file;
 /* Global variables */
 bool using_ha_suspend;
 
@@ -402,6 +403,23 @@ check_validate_config(void)
 	start_check(NULL, NULL);
 }
 
+/* Create the reload done file */
+int
+mark_reload_done(char *file_name)
+{
+	struct timeval t;
+    FILE *file = fopen(file_name, "w");
+
+	if (!file) {
+		log_message(LOG_INFO, "file_write : Can not open %s file",
+		       file_name);
+		return 0;
+	}
+	gettimeofday(&t, NULL);
+	fprintf(file, "reload done at %lu.%06lu\n", t.tv_sec, t.tv_usec);
+	fclose(file);
+	return 1;
+}
 #ifndef _DEBUG_
 /* Reload thread */
 static int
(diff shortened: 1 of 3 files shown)


ywc689:
What's the purpose of the created file on reload? It seems to record the reload finished time and not do anything else.
Besides, keepalived reload status can be obtained from the global var `reload`, which is set to be 1 when reloading and 0 any time else.

donghaobo:
> What's the purpose of the created file on reload? It seems to record the reload finished time and not do anything else.
> Besides, keepalived reload status can be obtained from the global var `reload`, which is set to be 1 when reloading and 0 any time else.

@ywc689  If we want to add vs in KEEPALIVED, we will update the configuration file and then use signal to trigger keepalived reload. If we want to add another vs, we don't know whether keepalived has completed the last reload. We want to update the configuration file, but keepalived may be reading the file.
If keepalived writes the mark file, we can prevent other reloading tasks and make the current reloading task wait until the reload is actually completed.

ywc689:
> > What's the purpose of the created file on reload? It seems to record the reload finished time and not do anything else.
> > Besides, keepalived reload status can be obtained from the global var `reload`, which is set to be 1 when reloading and 0 any time else.
> 
> @ywc689 If we want to add vs in KEEPALIVED, we will update the configuration file and then use signal to trigger keepalived reload. If we want to add another vs, we don't know whether keepalived has completed the last reload. We want to update the configuration file, but keepalived may be reading the file.
> If keepalived writes the mark file, we can prevent other reloading tasks and make the current reloading task wait until the reload is actually completed.

@donghaobo The codes seems just write a mark file each time reload. I'm wondering how can we prevent from other reloading tasks with the mark file? Is there any other dependent job process looking up the mark file and controlling the reload frequency?
Besides, what's the problem if a reload is triggered before previous reload has finished?

donghaobo:
> Besides, what's the problem if a reload is triggered before previous reload has finished?

@ywc689  
We reload in the following order:

1. use `flock` lock a file to prevent other reload
2. delete the mark file
3. do the reload
4. wait untile the mark file created
5. exit, `flock` auto released

when the last reloading is not complete, the keepalived is reading the config file, and the next reloading  write to the same config file. the reading is unpredictable. maybe you will get parts of the new config, and parts of old config. and things will be worse if it split an VS.
In fact, we have not encountered any problems for now.

ywc689:
As far as I can see, the mark file itself doesn't fix the reload reentrance problem, but rather as an ancillary tool for peripheral process which controls keepalived's reload. In terms of keepalived, the mark file is confusing, and a little abrupt. I think a better way  is to fix the problem within keepalived itself.

PR #669, opened 2020-10-27
software fdir & rss for traceroute

donghaobo:
use soft fdir rule to redirect ICMP Error packet, to support traceroute

### DIFF
--- a/include/rculist.h
+++ b/include/rculist.h
@@ -4,18 +4,59 @@
 #include <rte_atomic.h>
 #include "list.h"
 #define __rcu
-#define WRITE_ONCE(a, b) (a) = (b)
 #define rcu_assign_pointer(a, b) do {\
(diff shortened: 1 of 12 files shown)


include/rculist.h
```diff
@@ -0,0 +1,742 @@
+#ifndef _LINUX_RCULIST_H
+#define _LINUX_RCULIST_H
+
+#include <rte_atomic.h>
+#include "list.h"
+#define __rcu
+#define WRITE_ONCE(a, b) (a) = (b)
```
ywc689, line 7: Is the `WRITE_ONCE` safe enough for rculist? It's define in linux kernel source in other way. Refer to https://elixir.bootlin.com/linux/v4.18.20/source/include/linux/compiler.h#L273.

PR #674, opened 2020-11-09
main.c: code refactoring on the initialization/termination of all the…

andrewhit:
… modules.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -57,6 +57,94 @@ extern bool g_dpvs_pdump;
 #endif
 extern int log_slave_init(void);
 
+
+
+/*
+ * the initialization order of all the modules
+ */
+#define DPVS_MODULES {                                          \
+        DPVS_MODULE(MODULE_FIRST,       "scheduler",            \
+                    dpvs_scheduler_init, dpvs_scheduler_term),  \
+        DPVS_MODULE(MODULE_GLOBAL_DATA, "global data",          \
+                    global_data_init,    global_data_term),     \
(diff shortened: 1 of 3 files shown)


src/main.c
```diff
@@ -57,6 +57,96 @@ extern bool g_dpvs_pdump;
...
+            }
+        }
+    }
+}
+
+static void modules_term(void)
+{
+    int m, err;
+
+    for (m = MODULE_LAST; m >= MODULE_FIRST; m--) {
```
ywc689, line 140: the initial value for m should be `MODULE_LAST - 1`

src/main.c
```diff
@@ -57,6 +57,96 @@ extern bool g_dpvs_pdump;
...
+        if (dpvs_module_terms[m]) {
+            if ((err = dpvs_module_terms[m]()) != EDPVS_OK) {
```
ywc689, line 142: Some modules don't have cleanup functions. Check whether `dpvs_module_terms[m]` is NULL and call it only when not NULL.

ywc689 (changes requested):

src/main.c
andrewhit: Line 141 checks dpvs_module_terms[m], cleanup func is not-NULL.

src/main.c
andrewhit: Now the new patch makes "iftraf" as MODULE_LAST.

ywc689 (approved):

PR #676, opened 2020-11-09
fix the mlx5 config not work

legend050709:
fix the bug of CONFIG_MLX5=y in config.mk not work;

### DIFF
--- a/src/config.mk
+++ b/src/config.mk
@@ -47,8 +47,26 @@ CFLAGS += -D DPVS_MAX_LCORE=64
 
 ifeq ($(CONFIG_PDUMP), y)
 CFLAGS += -D CONFIG_DPVS_PDUMP
+LIBS += -Wl,--whole-archive -lrte_member -lrte_eventdev -lrte_reorder -lrte_cryptodev \
+		-lrte_vhost -lrte_pmd_pcap
+
+ifneq ("$(wildcard $(RTE_SDK)/$(RTE_TARGET)/lib/librte_bus_vmbus.a)", "")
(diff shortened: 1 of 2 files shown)


ywc689:
This bug in introduced by PR #667 . You could fixed it by exchanging the order of `include $(SRCDIR)/dpdk.mk` and `include $(SRCDIR)/config.mk` in src/Makefile.

legend050709:
> This bug is introduced by PR #667 . You could fixed it by exchanging the order of `include $(SRCDIR)/dpdk.mk` and `include $(SRCDIR)/config.mk` in src/Makefile.

legend050709:
> This bug is introduced by PR #667 . You could fixed it by exchanging the order of `include $(SRCDIR)/dpdk.mk` and `include $(SRCDIR)/config.mk` in src/Makefile.

Well, the $(RTE_SDK)/mk/rte.vars.mk included in dpdk.mk overrides CFLAGS, so we need to include dpdk.mk first. 
see the comment in [pr-667](https://github.com/iqiyi/dpvs/pull/667)

ywc689:
It's good practice to keep the dpdk make configs within the "src/dpdk.mk" file.  A more reasonable solution is to move the `include $(RTE_SDK)/mk/rte.vars.mk ` line  in "src/dpdk.mk" into "src/Makefile", in which placing it before all `include` statements.

ywc689:
Because the bug priority is high. I'v make another pr #686 to fix it. You can help review it. Thank you!

ywc689:
Fixed in pr #686 .

PR #684, opened 2020-11-18
Add port and proto match when deleting blacklist.

weiyanhua100:


src/ipvs/ip_vs_blklst.c
```diff
@@ -216,8 +216,10 @@ void dp_vs_blklst_flush(struct dp_vs_service *svc)
 
     for (hash = 0; hash < DPVS_BLKLST_TAB_SIZE; hash++) {
         list_for_each_entry_safe(entry, next, &this_blklst_tab[hash], list) {
-            if (entry->af == svc->af &&
-                    inet_addr_equal(svc->af, &entry->vaddr, &svc->addr))
+            if (entry->af == svc->af
+                && entry->port == svc->port
```
ywc689, line 220: `entry` doesn't have member `port`. I think it won't compile.

weiyanhua100: sorry, port does not exist, it has been modified.

ywc689 (approved):

PR #687, opened 2020-11-23
Fix DEBUG level do not work with dpdk-stable-18.11.2

happylixiao:
Because dpdk-stable-18.11.2 rte_logs.dynamic_types default level is "INFO".

### DIFF
--- a/src/global_conf.c
+++ b/src/global_conf.c
@@ -33,9 +33,10 @@ static void log_current_time(void)
 
 static int set_log_level(char *log_level)
 {
-    if (!log_level)
+    if (!log_level) {
         rte_log_set_global_level(RTE_LOG_DEBUG);
-    else if (!strncmp(log_level, "EMERG", strlen("EMERG")))
+        rte_log_set_level_regexp("user[0-9]", RTE_LOG_DEBUG);
+    } else if (!strncmp(log_level, "EMERG", strlen("EMERG")))
         rte_log_set_global_level(RTE_LOG_EMERG);
     else if (!strncmp(log_level, "ALERT", strlen("ALERT")))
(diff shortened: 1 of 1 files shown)


src/global_conf.c
```diff
@@ -49,9 +50,10 @@ static int set_log_level(char *log_level)
         rte_log_set_global_level(RTE_LOG_NOTICE);
     else if (!strncmp(log_level, "INFO", strlen("INFO")))
         rte_log_set_global_level(RTE_LOG_INFO);
-    else if (!strncmp(log_level, "DEBUG", strlen("DEBUG")))
+    else if (!strncmp(log_level, "DEBUG", strlen("DEBUG"))) {
         rte_log_set_global_level(RTE_LOG_DEBUG);
-    else {
+        rte_log_set_level(RTE_LOGTYPE_USER1, RTE_LOG_DEBUG);
+    } else {
```
ywc689, line 56: Besides `RTE_LOGTYPE_USER1`, `RTE_LOGTYPE_USER2` and `RTE_LOGTYPE_USER3` are also used by dpvs.
Maybe a more feasible solution is to use `rte_log_set_level_pattern` instead.
```
rte_log_set_level_pattern("user[0-9]", RTE_LOG_DEBUG)
```

ywc689 (approved):

ywc689:
`rte_log_set_level_pattern` is not defined until dpdk-18.05.  The patch won't compile with dpdk-17.xx. Fix it using macro `RTE_VERSION_NUM` to support early  dpdk versions.

```
#if RTE_VERSION >= RTE_VERSION_NUM(17, 5, 0, 0)
...
#endif
```

PR #689, opened 2020-12-01
keepalived: fix epoll_wait return err

legend050709:
**Describe the bug**
keepalived running too long，which lead to print “scheduler：epoll_wait error ：invalid argument”

**To Reproduce**

add same vs or rs to keepalived.conf
systemctl start keepalived
wait some time。

gdb att checker process.
watch master-> epoll_count

**Expected behavior**
when keepalived.conf is not changed, the epoll_count is not increased。
under a lot of checkers, donot print "scheduler：epoll_wait error ：invalid argument" （because epoll_count is reversed）

see: https://github.com/acassen/keepalived/issues/1631

### DIFF
--- a/tools/keepalived/lib/scheduler.c
+++ b/tools/keepalived/lib/scheduler.c
@@ -616,6 +616,7 @@ thread_event_cancel(const thread_t *thread_cp)
 	rb_erase(&event->n, &m->io_events);
 	if (event == m->current_event)
 		m->current_event = NULL;
+	thread_events_resize(m, -1);
 	FREE(thread->event);
 	return 0;
 }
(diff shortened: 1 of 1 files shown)


tools/keepalived/lib/scheduler.c
```diff
@@ -1641,15 +1642,15 @@ thread_fetch_next_queue(thread_master_t *m)
 
 	assert(m != NULL);
 
-	/* If there is event process it first. */
```
ytwang0320, line 1644: Please confirm, @acassen/keepalived#1631 just added "thread_events_resize" but did not change this loop.

legend050709: The change of loop is for optimization. 
For example, the checker child process runs in the loop of thread_fetch_next_queue， then we send HUP signal to keepalived for reloading, epoll_wait  will return EINVAL error, and then it will create an THREAD_EVENT type event by the Signal callback function of thread_add_event. The reload operation will not be executed immediately, and will not be executed until the loop ends and enters the function again;

ytwang0320 (approved):

ywc689 (approved):

PR #690, opened 2020-12-07
support RS in vxlan network

donghaobo:
need pr #669


haosdent:
@donghaobo just curious, do you use this in production?

PR #696, opened 2021-01-04
pdump: fix pdump initialization problem introduced by pr #674

ywc689:
Fix pdump initialization problem introduced by pr #674.
Pdump should be initialized after cfgfile module, or `g_dpvs_pdump` is not set before used.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -46,6 +46,7 @@
 #include "iftraf.h"
 #include "eal_mem.h"
 #include "scheduler.h"
+#include "pdump.h"
 
 #define DPVS    "dpvs"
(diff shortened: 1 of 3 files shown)


src/main.c
```diff
@@ -69,6 +70,8 @@ extern int log_slave_init(void);
                     global_data_init,    global_data_term),     \
         DPVS_MODULE(MODULE_CFG,         "config file",          \
                     cfgfile_init,        cfgfile_term),         \
+        DPVS_MODULE(MODULE_PDUMP,        "pdump",               \
+                    pdump_init,          pdump_term),           \
```
yifan1227, line 74: The last module's name should be "eal_mem"

yifan1227 (approved):

PR #699, opened 2021-01-11
v1.8.8

ywc689:
Merge `devel` to `master` branch, and release v1.8.8.


you-looks-not-tasty (approved):

README.md
```diff
@@ -190,7 +192,7 @@ If you see this message. Well done, `DPVS` is working with NIC `dpdk0`!
 EAL: Error - exiting with code: 1
   Cause: ports in DPDK RTE (2) != ports in dpvs.conf(1)
 ```
->It means the NIC used by DPVS do not match `/etc/dpvs.conf`. Please use `dpdk-devbind` to adjust the NIC number or modify `dpvs.conf`. We'll improve this part to make DPVS more "clever" to avoid modify config file when NIC count do not match.
+>It means the NIC count of DPVS does not match `/etc/dpvs.conf`. Please use `dpdk-devbind` to adjust the NIC number or modify `dpvs.conf`. We'll improve this part to make DPVS more "clever" to avoid modify config file when NIC count does not match.
```
yifan1227, line 195: It means **that** the NIC count of DPVS does not match **that in** `/etc/dpvs.conf`. Please use `dpdk-devbind` to adjust the NIC number or modify `dpvs.conf`. We'll improve this part to make DPVS more "clever" to avoid **modifying the** config file **when the numbers of NIC** do not match.

yifan1227 (approved):

PR #720, opened 2021-03-10
TC per-lcore tuning

ywc689:
- Refactor TC module to be per-lcore.
- Fix crash problem when add/modify/del qsch/qcls dynamically with traffic bursting in.
- Fix inaccuracy traffic control problem of tbf caused by multi-thread contention of tokens.
- Schedule qsch per loop to make TC work properly.


haosdent:
Would this make QoS usable?

ywc689:
> Would this make QoS usable?

Just some simple cases, such as traffic shaping and policing. I'll write a doc to show some use cases later.

haosdent:
I think, thanks a lot! @ywc689

include/linux_ipv6.h
```diff
@@ -266,6 +266,18 @@ static inline int ipv6_addr_cmp(const struct in6_addr *a1, const struct in6_addr
     return memcmp(a1, a2, sizeof(struct in6_addr));
 }
 
+static inline int ipv6_addr_cmp_u128(const struct in6_addr *a1, const struct in6_addr *a2)
```
yifan1227, line 269: What's the difference between thiss function and the upper one?

yifan1227, line 269: This function is duplicated with the upper one.

yifan1227 (approved):

PR #739, opened 2021-07-02
feature: graceful exit.

zhangtengfei-oppo:
expire all conns before exit.
send rst to all tcp endpoints.

### DIFF
--- a/src/signals.c
+++ b/src/signals.c
@@ -1,0 +1,111 @@
+#include "parser/parser.h"
+#include "ipvs/conn.h"
+#include "ctrl.h"
+
+#include <signal.h>
+
+#define RTE_LOGTYPE_SIGNALS     RTE_LOGTYPE_USER1
+
+static int signal_terminate_conns(struct dpvs_msg* msg);
+
+static struct dpvs_msg_type signal_msg_type = {
+        .type               = MSG_TYPE_SIG_TERM_CONNS,
+        .prio               = MSG_PRIO_HIGH,
+        .mode               = DPVS_MSG_MULTICAST,
+        .unicast_msg_cb     = signal_terminate_conns,
+        .multicast_msg_cb   = NULL,
+};
+
+static int signal_terminate_conns(struct dpvs_msg* msg) 
+{
+
+    dp_vs_conn_term_all();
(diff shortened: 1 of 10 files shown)


src/signals.c
```diff
@@ -0,0 +1,111 @@
...
+        .unicast_msg_cb     = signal_terminate_conns,
+        .multicast_msg_cb   = NULL,
+};
+
+static int signal_terminate_conns(struct dpvs_msg* msg) 
+{
+
+    dp_vs_conn_term_all();
+    //flush pkts in hardware cache
+    netif_hard_flush_by_lcore(rte_lcore_id());
```
ywc689, line 24: Packets cannot be sent out after connection table being flushed.

ywc689:
1. Could you give a case where the graceful exit to use?
2. I don't think "send rst to all tcp endpoints" is done in this patch.
3. Does the patch interfere with dpvs coredump or gdb debug?

src/signals.c
zhangtengfei-oppo: > 1. Could you give a case where the graceful exit to use?
> 2. I don't think "send rst to all tcp endpoints" is done in this patch.
> 3. Does the patch interfere with dpvs coredump or gdb debug?

1、In a dpvs cluster, the same virtual IP is distributed on multiple machines. If a machine is going offline, because there is no session synchronization mechanism, the TCP long connection will fail over time. In order to avoid TCP long connection timeout and reconnection, it should be closed immediately before exiting. 

2、the  function "dp_vs_conn_term_all" will traverse the connection table , and call  "tcp_conn_expire". it will send rst to RS and client

3、gdb breakpoint use SIGTRAP，coredump use many signals, most of them can't be catched by  'sigaction' except SIGQUIT, this patch catch  SIGINT、SIGTERM、SIGQUIT.  SIGQUIT was send by termainl when  typing keys "Ctrl-\“ , This signal  basically won't happen ，because dpvs run in background. Of course, it’s okay to delete this signal

zhangtengfei-oppo: > Packets cannot be sent out after connection table being flushed.

I didn't understand what you mean

PR #768, opened 2021-11-01
Ipset Update

ywc689:


src/ipset/ipset_hash.c
```diff
@@ -167,23 +166,26 @@ hash_del(struct ipset *set, void *value, uint16_t flag)
...
     struct hash_entry *hnode;
     struct list_head *head ;
 
     key = do(hash, elem, set->hash_len, htype->mask);
     head = &htype->htable[key];
     list_for_each_entry(hnode, head, list) {
-        if (do(compare, elem, hnode->elem))
-            return 1;
+        res = do(compare, elem, hnode->elem);
+        if (res == COMPARE_EQUAL_ACCEPT ||
```
Geek-fan, line 178: Is it ok to use `if (res != COMPARE_INEQUAL)`

doc/IPset.md
```diff
@@ -2,218 +2,709 @@ DPVS IPset
...
+    NET       := "{ IP/prefix | IP(range from)[-IP(range end)] }"
+    MAC       := 6 bytes MAC address string literal
+    PORT      := "[{ tcp | udp | icmp | icmp6 }:]port1[-port2]"
+    OPTIONS   := { comment | range NET | hashsize NUM | maxelem NUM }
+    ADTOPTS   := { comment STRING | unmatch (for add only) }
+    flag      := { -F(--force) | { -4 | -6 } | -v }
 ```
 
-<a id='examples'/>
+For example, the following command creates a hash:net,port type ipset named `foo`, whose hash table bucket size is 256, and can store 1000 entries at most.
```
Geek-fan, line 55: ... 1000 entries at most, with comment enabled.

doc/IPset.md
```diff
@@ -2,218 +2,709 @@ DPVS IPset
...
-# dpip ipset add foo 192.168.100.100,AA:BB:CC:DD:EE:FF comment "initial"
-# dpip ipset add foo 192.168.100.100,AA:BB:CC:DD:EE:FF comment "overwrite" -F
-# dpip ipset list
+
+Then we add some entries into the ipset foo.
+
+```bash
+ ./bin/dpip ipset add foo 10.132.0.0/16,udp:10240
+ ./bin/dpip ipset add foo 192.168.88.0/24,tcp:8080-8082
+ ./bin/dpip ipset add foo 192.168.88.200-192.168.88.255,tcp:8082 nomatch
```
Geek-fan, line 65: ...tcp:8082 nomatch comment "bad guys"

doc/IPset.md
```diff
@@ -2,218 +2,709 @@ DPVS IPset
...
+192.168.88.208/28,tcp:8082 nomatch  comment "bad guys"
+192.168.88.224/27,tcp:8082 nomatch  comment "bad guys"
+192.168.88.0/24,tcp:8080  
+192.168.88.0/24,tcp:8081  
+192.168.88.0/24,tcp:8082  
+10.132.0.0/16,udp:10240
+```
+Note that 7 ipset entries are created, and IPv4 range is transformed to CIDR range format when stored into ipset. The flag `-v` in the command above indicates to sort ipset entries if the ipset type supports `sort_compare` method in dpip.
+
+Finally, let's make some tests.
```
Geek-fan, line 91: let's do some tests

doc/IPset.md
```diff
@@ -2,218 +2,709 @@ DPVS IPset
...
+10.100.100.100,udp:10240 is NOT in set foo
+# ./bin/dpip ipset test foo 192.168.88.0,tcp:8080
+true
+# ./bin/dpip ipset test foo 192.168.88.22,tcp:8082
+true
+# ./bin/dpip ipset test foo 192.168.88.222,tcp:8082
+false
+```
+The last test returned "false" because of the nomatch entry `192.168.88.200-192.168.88.255,tcp:8082 nomatch`. If we add one more specific entry `192.168.88.222/32,tcp:8082` and test again, the result would turn out to be "true".
+
```
Geek-fan, line 106: Should we explain that entry with larger CIDR number have higher priority?

doc/IPset.md
```diff
@@ -2,218 +2,709 @@ DPVS IPset
...
+# Set Types
+
+> For more exmaple of ipset types, refer to [ipset test script](../test/ipset/dpip.sh) and the [test results](../test/ipset/dpip.log).
+
+<a id='bitmap_ip'/>
+
+#### bitmap:ip
+
+The bitmap:ip set type uses a memory range to store IPv4 addresses. IPv4 range is supported when add/delete entries, but the ipset parses the range and finds all specific IPv4 addresses in the range, then save them in the bitmap storage. Therefore, we can add an IPv4 range to bitmap:ip, but delete a subrange IPv4 address from the ipset. IPv6 is not supported by the type.
```
Geek-fan, line 140: IPv4 range is supported when add/delete entries: ipset parses the range and derives all specific IPv4 addresses in the range, then save them in the bitmap storage.

doc/IPset.md
```diff
@@ -2,218 +2,709 @@ DPVS IPset
...
+192.168.1.2  
+192.168.1.3  
+192.168.1.7 
+```
+
+<a id='bitmap_ip_mac'/>
+
+#### bitmap:ip,mac
+
+The bitmap:ip,mac set type uses a memory range to store IPv4 and a MAC address pairs. If MAC address is not specified with a entry, then MAC address is ignored when match against this entry, i.e., only IPv4 address is considered in this case.
```
Geek-fan, line 181: If MAC address is not specified in the pair, it is ignored when matching against this entry,

doc/IPset.md
```diff
@@ -2,218 +2,709 @@ DPVS IPset
...
+bin/dpip: invalid parameter
+```
+
+<a id='hash_ip_port_net'/>
+
+#### hash:ip,port,net
+
+The hash:ip,port,net set type uses a hash table to store IP address, port number and IP network address triples. Both IPv4 and IPv6 address are supported. The IP address of the IP and net should be of the same family. When adding/deleting entries, ranges are allowed but is transformed to specific host IP and port entries when stored into hash table for the "ip" and "port" segments. IPv4 supports both IP range and IP CIDR, while IPv6 supports IP CIDR only. Network address with zero prefix size is not supported, and is interpreted as host prefix size, i.e., 32 for IPv4 and 128 for IPv6. Option "nomatch" can be used to set exceptions to the set when add/deleting entries. If a test is matched against with a "nomatch" entry, then the result would end with false. The port number is interpreted together with a protocol. Supported protocols include TCP, UDP, ICMP, and ICMPv6, any other protocols are interpreted as unspec type with a protocol number of zero.
+
+```bash
```
Geek-fan, line 461: `# ./bin/dpip ipset create bar hash:ip,port,net
 # ./bin/dpip ipset add bar 2001::1,8080-8082,2002::/64` 
 are missed

doc/IPset.md
```diff
@@ -2,218 +2,709 @@ DPVS IPset
...
+# ./bin/dpip ipset -v show bar
+Name: bar
+Type: hash:net,port,net,port
+Header: family inet6  hashsize 1024  maxelem 65535  comment
+Size in memory: 17296
+References: 0
+Number of entries: 6
+Members:
+2001::/64,udp:8080,2002::1:2:0:0/96,udp:6000 nomatch  comment "bad-guys"
+2001::/64,udp:8080,2002::1:2:0:0/96,udp:6001 nomatch  comment "bad-guys" 2001::/64,udp:8080,2002::/64,udp:6000
```
Geek-fan, line 563: should be in a new line

doc/IPset.md
```diff
@@ -2,218 +2,709 @@ DPVS IPset
...
+10.60.100.112/28,tcp:10242,10.130.100.0/24 nomatch  
+10.60.100.128/26,tcp:10242,10.130.100.0/24 nomatch  
+10.60.0.0/16,tcp:10240,10.130.0.0/16  
+10.60.0.0/16,tcp:10241,10.130.0.0/16  
+10.60.0.0/16,tcp:10242,10.130.0.0/16
+# ./bin/dpip ipset test foo 10.60.0.0,tcp:10242,10.130.255.255
+true
+# ./bin/dpip ipset test foo 10.60.100.111,tcp:10240,10.130.3.4
+true
+# ./bin/dpip ipset test foo 10.60.100.111,tcp:10240,10.130.100.222
```
Geek-fan, line 516: Is there an example interpreting "The first "net" segment takes precedence over the second one when matching against the set, regardless of whether the entry is of "nomatch" option"?

src/ipset/ipset_hash_ipportnet.c
```diff
@@ -0,0 +1,245 @@
...
+    if (param->cidr2) {
+        ip_set_mask_from_to(ip2_from, ip2_to, param->cidr2);
+    } else {
+        ip2_to = ntohl(param->range2.max_addr.in.s_addr);
+    }
+
+    do {
+        e.ip.in.s_addr = htonl(ip);
+        ip = ip_set_range_to_cidr(ip, ip_to, &e.cidr);
+        for (ip2 = ip2_from; ip2 <= ip2_to; ip2++) {
```
Geek-fan, line 120: Should ip and ip2 swap position to avoid confusing? Because this method is called IP-Port-Net.

tools/dpip/ipset.c
```diff
@@ -382,85 +530,285 @@ get_type_idx(char *type)
...
+        }
+        if (ntohl(param.range.max_addr.in.s_addr) != 0) {
+            if (ntohl(param.range.max_addr.in.s_addr) <
+                    ntohl(param.range.min_addr.in.s_addr)) {
+                fprintf(stderr, "invalid ipv4 range\n");
+                return EDPVS_INVAL;
+            }
+            if (ntohl(param.range.max_addr.in.s_addr) -
+                    ntohl(param.range.min_addr.in.s_addr) > 65536) {
+                fprintf(stderr, "ipv6 range shouldn't be greater than 65536\n");
```
Geek-fan, line 700: Is here supposed to be ip range?

tools/dpip/ipset.c
```diff
@@ -382,85 +530,285 @@ get_type_idx(char *type)
...
+                return EDPVS_INVAL;
+            }
+        }
+        if (ntohl(param.range2.max_addr.in.s_addr) != 0) {
+            if (ntohl(param.range2.max_addr.in.s_addr) <
+                    ntohl(param.range2.min_addr.in.s_addr)) {
+                fprintf(stderr, "invalid ipv4 range\n");
```
Geek-fan, line 707: I think it's checked when parsing. (line 188)

tools/dpip/ipset.c
```diff
@@ -143,37 +196,52 @@ addr_arg_parse(char *arg, struct inet_addr_range *range, uint8_t *cidr)
...
     if ((sep = strchr(arg, ':')) != NULL) {
         *sep++ = '\0';
         arg = sep;
-        if (!strcmp(proto, "udp")) 
-            param.proto = IPPROTO_UDP;
     }
 
     port1 = arg;
     range->max_port = range->min_port = atoi(port1);
+    if (range->min_port > 65535)
```
Geek-fan, line 223: The type of min_port/max_port is __be16. Is this check necessary?

Geek-fan (commented):
See comments.

src/ipset/ipset_hash.c
ywc689: yes. but it's easier to understand in this way.

doc/IPset.md
ywc689: add an example:

```
# ./bin/dpip ipset flush foo 
# ./bin/dpip ipset add foo 192.168.100.0/24,udp:6000,192.168.200.0/25
# ./bin/dpip ipset add foo 192.168.100.0/25,udp:6000,192.168.200.0/24 nomatch
# ./bin/dpip ipset list foo 
Name: foo 
Type: hash:net,port,net
Header: family inet  hashsize 1024  maxelem 65535  
Size in memory: 16688
References: 0
Number of entries: 2
Members:
192.168.100.0/24,udp:6000,192.168.200.0/25  
192.168.100.0/25,udp:6000,192.168.200.0/24 nomatch  
# ./bin/dpip ipset test foo 192.168.100.1,udp:6000,192.168.200.1
false
# ./bin/dpip ipset flush foo
# ./bin/dpip ipset add foo 192.168.100.0/24,udp:6000,192.168.200.0/25 nomatch
# ./bin/dpip ipset add foo 192.168.100.0/25,udp:6000,192.168.200.0/24
# ./bin/dpip ipset list foo
Name: foo
Type: hash:net,port,net
Header: family inet  hashsize 1024  maxelem 65535
Size in memory: 16688
References: 0
Number of entries: 2
Members:
192.168.100.0/24,udp:6000,192.168.200.0/25 nomatch
192.168.100.0/25,udp:6000,192.168.200.0/24
[root@10.15.204.12 dpvs]# ./bin/dpip ipset test foo 192.168.100.1,udp:6000,192.168.200.1
true
```

src/ipset/ipset_hash_ipportnet.c
ywc689: It's confined by the basic `hash_elem` memory arrangement. If a cidr not used, it should be cidr2.

```
typedef struct hash_elem {
    union inet_addr ip1;
    uint8_t cidr;
    union inet_addr ip2;
    uint8_t cidr2;
} elem_t;
```

tools/dpip/ipset.c
ywc689: yes.

tools/dpip/ipset.c
ywc689: fixed

PR #771, opened 2021-11-10
A new TC cls based on IPset

yfgogogo:


src/tc/cls_ipset.c
```diff
@@ -0,0 +1,192 @@
...
+        offset += VLAN_HLEN;
+        goto l2parse;
+
+    default:
+        goto done;
+    }
+
+    direction = cls->sch->flags & QSCH_F_INGRESS? 1 : -1;
+    /* for IPset type containing MAC element */
+    mbuf->userdata = direction == 1?
```
Geek-fan, line 115: This can be optimized in new DPDK version.

ywc689:
refer to pr #776

PR #793, opened 2022-01-06
modify bug: icmp6 mtu error, tunnel mode icmp/icmp6 error

yfming:
tunnel模式下，报文添加外层ip头，长度超过mtu后，返回的icm包里mtu值应该减去外层ip头长度。
icmp6的构造传入参数应该是主机序列的，因为icmp6_send函数里面对传入的参数mtu做了大端转换。

### DIFF
--- a/src/ipvs/ip_vs_core.c
+++ b/src/ipvs/ip_vs_core.c
@@ -502,7 +502,7 @@ static int __xmit_outbound_icmp6(struct rte_mbuf *mbuf,
 
     if (mbuf->pkt_len > rt6->rt6_mtu) {
         route6_put(rt6);
-        icmp6_send(mbuf, ICMP6_PACKET_TOO_BIG, 0, htonl(rt6->rt6_mtu));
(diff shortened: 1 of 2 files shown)


ywc689 (approved):

ywc689:
Good work! 
 But I tested that ICMP_UNREACH_NEEDFRAG is not sent out when mtu is not safisfied.  It's caused by the empty route cache in `icmp_send`. I hope that you can fix it later.

yfming:
> Good work! But I tested that ICMP_UNREACH_NEEDFRAG is not sent out when mtu is not safisfied. It's caused by the empty route cache in `icmp_send`. I hope that you can fix it later.

是哪种模式呢？

ywc689:
> Good work! But I tested that ICMP_UNREACH_NEEDFRAG is not sent out when mtu is not safisfied. It's caused by the empty route cache in `icmp_send`. I hope that you can fix it later.

fixed in  #810

PR #801, opened 2022-02-08
bugfix: ipv4 fragmentation causes crash

yfming:


ytwang0320 (approved):
This bug introduced when dpdk-stable-20.11 merged.

PR #805, opened 2022-02-16
Uoa in container

ywc689:
The patch fixed the following problems.

1. In opp mode, after uoa data parsed, the opp header and data are translated to ipv4 header NOOP options, where the memory of packets is not dealt with correctly.
2. Some container network enforcements may hinder the host from passing to container the packets with uoa private protocol data, that is ip protocol 254. Thus add an optional param to uoa to parse uoa data at the tail of netfilter forwarding chain.  Though it's not  a thorough solution due to the complixity and diversity of container networking enforcements, we verified that it does work with k8s cilium cni deployed without encapsultion.


ytwang0320 (approved):
+1

PR #835, opened 2022-06-09
synproxy: fix fullnat tcp forwarding problem when defer_rs_syn enabled

ywc689:
Tcp reciece window of the client is closed in the first step of synproxy,
and client could only send window probe ack packets without payload in the
case. So it conflicts with the defer_rs_syn, which schedules ipvs connection
only when tcp ack packets with payload catched in the second step of synproxy.

Besides, a config item close_client_window is added to disable client window
close in the first step of synproxy.

Signed-off-by: ywc689 <ywc689@163.com>

### DIFF
--- a/conf/dpvs.conf.items
+++ b/conf/dpvs.conf.items
@@ -260,6 +260,7 @@ ipvs_defs {
             !    wscale                      <disable>
             !    timestamp                   <disable>
             }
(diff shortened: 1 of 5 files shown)


ytwang0320 (approved):
approve +1

PR #869, opened 2023-02-08
urgencty bugfix about black&white list 

you-looks-not-tasty:
set black or white addr to correct object member


ywc689 (approved):

ywc689:
The static ip entry deletion seems still not working.

PR #884, opened 2023-04-18
ipvs: backend health check

ywc689:
It works only for TCP by checking if syn-ack is replied from the backend, and if not, inhibit scheduling the backend for a period. The inhibiting period is gained or lost exponentially depending on successive checking is failed or succeeded.

UDP dest-check works only when the udp flow is bidirectional, that is, the udp conns that forwarding inbound only or outbound only are always detcted as dead. Thus "dest-check" should be never configured for the single directional udp flow. Besides, the a smaller conn timeout may be specified for the bidirectional flow service to detect dest fault quickly.


ytwang0320 (approved):
approve +1

PR #891, opened 2023-05-16
keepalived: don't delete and re-add vs/rs when update at reload

roykingz:
when update vs/rs configuration parameters at keepalived reload, it's no need to delete and re-add vs/rs, which will result in traffic interruption of the vs/rs.


ywc689 (approved):

ywc689:
I tested that the vs's protocol was not changed when modified. For example, change the protocol from TCP to UDP and reload keepalived, but the effective protocol still is TCP.

ywc689:
> I tested that the vs's protocol was not changed when modified. For example, change the protocol from TCP to UDP and reload keepalived, but the effective protocol still is TCP.

fixed by 6652d7e94914e49abf30c45fecb2846a851a1a69

PR #901, opened 2023-07-28
timeout > 524 is not accurate

donghaobo:


src/timer.c
```diff
@@ -187,16 +187,19 @@ static int __dpvs_timer_sched(struct timer_scheduler *sched,
...
-        if (off > 0) {
+        if (off < LEVEL_SIZE) {
             hash = (sched->cursors[level] + off) % LEVEL_SIZE;
             list_add_tail(&timer->list, &sched->hashs[level][hash]);
 #ifdef CONFIG_TIMER_DEBUG
             assert(timer->handler == handler);
 #endif
             return EDPVS_OK;
+        } else {
+            timer->delay += sched->cursors[level] * get_level_ticks(level);
```
ywc689, line 202: `timer->delay` is an user-specified parameter, and we should not changed. It seems a carry-in problem. You may fix it in some other way.

donghaobo: > `timer->delay` is an user-specified parameter, and we should not changed. It seems a carry-in problem. You may fix it in some other way.

seems #29  fixed, but not merged

ywc689: Yes. I will reconsider it later.

PR #912, opened 2023-09-14
added dcdn toa

ytwang0320:


src/ipvs/ip_vs_proto_tcp.c
```diff
@@ -305,6 +305,43 @@ static void tcp_in_remove_ts(struct tcphdr *tcph)
...
+            opsize = *ptr++;
+            if (opsize < 2)    /* silly options */
+                return EDPVS_NOTEXIST;
+            if (opsize > len)
+                return EDPVS_NOTEXIST;    /* partial options */
+            if ((opcode == TCP_OPT_DCDN_ADDR)
+                    && (*ptr == TOA_DCDN_IPV4) && (opsize == TCP_OLEN_IP4_ADDR - 1) && addr) {
+                memcpy(addr, ptr + 1, sizeof(struct in_addr));
+                return EDPVS_OK;
+            }
```
ywc689, line 337: The client's TCP port isn't considered, why?

src/ipvs/ip_vs_proto_tcp.c
```diff
@@ -720,6 +764,14 @@ static int tcp_fnat_in_handler(struct dp_vs_proto *proto,
     if (th->syn && !th->ack) {
         tcp_in_remove_ts(th);
         tcp_in_init_seq(conn, mbuf, th);
+        if (tcp_in_check_toa(conn, th, &dcdn_addr) == EDPVS_OK) {
+            conn->dcdn_found = true;
+            conn->dcdn_addr = dcdn_addr;
```
ywc689, line 769: It seems every TCP packet go through the `tcp_in_check_toa`. It may be optimized with the `dcdn_found` flag by checking the toa only when the flag is not set.

ywc689 (commented):
This feature is a special use case for aliyun, and not expected to get into the DPVS mainline.
It's advised to submit a patch file instead.

PR #920, opened 2023-10-20
fix rs setting racing problem between dpvs-agent and healthcheck

you-looks-not-tasty:
Healthcheck module set rs.Weight would occurred some problems at the gaps of adjacent consistent rs spec info.

The API check if rs.consistentWeight equal to dpvs active weight.
rs.consistentWeight != dpvs active weight, HTTP STATUS CODE = 270
rs.consistentWeight == dpvs active weight, HTTP STATUS CODE = 200
And responed the currently valid value of rs.Weight to client(dpvs-healthcheck).


ywc689:
track the problem in #937 , closing this one.

PR #928, opened 2023-12-11
ipvs: toa enhancements

ywc689:
1. Do not insert toa data in syn packets any more.
2. Compact tcp option space or remove some options when inserting toa data failed  and then try again.
3. Emit a warning log for audit when toa insertion failed finally.

### DIFF
--- a/src/ipvs/ip_vs_proto_tcp.c
+++ b/src/ipvs/ip_vs_proto_tcp.c
@@ -306,46 +306,129 @@ static void tcp_in_remove_ts(struct tcphdr *tcph)
     }
 }
 
-/* use NOP option to replace TCP_OLEN_IP4_ADDR and TCP_OLEN_IP6_ADDR opt */
-static void tcp_in_remove_toa(struct tcphdr *tcph, int af)
+/*
+ * Remove NOP and TOA options preset in the mbuf and compact option space.
+ * If still no enough space, trim more options except for the protected ones.
+ *
+ * Return the trimmed length on success, otherwise dpvs error num on failure.
+ * */
+static int tcp_in_prune_options(int af, int reqlen, struct rte_mbuf *mbuf, struct tcphdr *tcph)
 {
-    unsigned char *ptr;
-    int len, i;
-    uint32_t tcp_opt_len = af == AF_INET ? TCP_OLEN_IP4_ADDR : TCP_OLEN_IP6_ADDR;
+    unsigned char *ptr, *fast, *slow;
+    const unsigned char *l3hdr, *payload;
+    int i, optlen;
+    unsigned int pruned;
+    uint8_t opcode, opsize;
+    uint64_t opts_protected;
+    const uint8_t opts_maxlen[64] = {
+        [2] = 4,  [3] = 3,   [4] = 2,
+        [8] = 10, [30] = 40, [34] = 18
+    };
 
     ptr = (unsigned char *)(tcph + 1);
-    len = (tcph->doff << 2) - sizeof(struct tcphdr);
+    fast = slow = ptr;
+    optlen = (tcph->doff << 2) - sizeof(struct tcphdr);
+    payload = ptr + optlen;
 
-    while (len > 0) {
-        int opcode = *ptr++;
-        int opsize;
+    if (optlen < reqlen)    /* make no sense to do anything */
+        return 0;
 
+    while (optlen > 0) {
+        opcode = *ptr++;
         switch (opcode) {
         case TCP_OPT_EOL:
-            return;
+            goto fini;
         case TCP_OPT_NOP:
-            len--;
+            fast++;
+            optlen--;
             continue;
         default:
             opsize = *ptr++;
-            if (opsize < 2)    /* silly options */
-                return;
-            if (opsize > len)
-                return;    /* partial options */
-            if ((opcode == TCP_OPT_ADDR) && (opsize == tcp_opt_len)) {
-                for (i = 0; i < tcp_opt_len; i++) {
-                    *(ptr - 2 + i) = TCP_OPT_NOP;
+            if (opsize < 2)             /* silly options */
+                goto fini;
+            if (opsize > optlen)        /* partial options */
+                goto fini;
+            if (opcode == TCP_OPT_ADDR) {
+                fast += opsize;
+            } else {
+                for (i = 0; i < opsize; i++) {
+                    if (slow != fast)
+                        *slow = *fast;
+                    slow++;
+                    fast++;
                 }
-                /* DON'T RETURN
-                 * keep search other TCP_OPT_ADDR ,and clear them.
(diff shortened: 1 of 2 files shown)


ytwang0320 (approved):
approve +1

src/ipvs/ip_vs_proto_tcp.c
```diff
@@ -306,46 +306,88 @@ static void tcp_in_remove_ts(struct tcphdr *tcph)
...
+    fast = slow = ptr;
+    optlen = (tcph->doff << 2) - sizeof(struct tcphdr);
+    payload = ptr + optlen;
 
+    while (optlen > 0) {
+        opcode = *ptr++;
         switch (opcode) {
         case TCP_OPT_EOL:
-            return;
+            goto fini;
```
donghaobo, line 330: set `fast += optlen` to trim unused options?

ywc689: `payload` , a pointer to the beginning of l4 payload, is used instead of the `fast` pointer when trimming the packet. I think it's an equivalent to what you mean, and in a safer way.

src/ipvs/ip_vs_proto_tcp.c
```diff
@@ -306,46 +306,88 @@ static void tcp_in_remove_ts(struct tcphdr *tcph)
...
             opsize = *ptr++;
-            if (opsize < 2)    /* silly options */
-                return;
-            if (opsize > len)
-                return;    /* partial options */
-            if ((opcode == TCP_OPT_ADDR) && (opsize == tcp_opt_len)) {
-                for (i = 0; i < tcp_opt_len; i++) {
-                    *(ptr - 2 + i) = TCP_OPT_NOP;
+            if (opsize < 2)             /* silly options */
+                return EDPVS_INVPKT;
```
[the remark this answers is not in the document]
ywc689: `opsize` is used in the loop condition later. I'm not sure how to deal with it if not valid. Removing all the tcp options?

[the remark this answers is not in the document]
donghaobo: > `opsize` is used in the loop condition later. I'm not sure how to deal with it if not valid. Removing all the tcp options?

I recommend removing all later tcp options.
the linux kernel will alse stop parsing later options.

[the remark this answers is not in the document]
cfc4n: When the `opcode` is `TCP_OPT_NOP`, it can be removed

src/ipvs/ip_vs_proto_tcp.c
```diff
@@ -306,46 +306,129 @@ static void tcp_in_remove_ts(struct tcphdr *tcph)
...
-                return;    /* partial options */
-            if ((opcode == TCP_OPT_ADDR) && (opsize == tcp_opt_len)) {
-                for (i = 0; i < tcp_opt_len; i++) {
-                    *(ptr - 2 + i) = TCP_OPT_NOP;
+            if (opsize < 2)             /* silly options */
+                goto fini;
+            if (opsize > optlen)        /* partial options */
+                goto fini;
+            if (opcode == TCP_OPT_ADDR) {
+                fast += opsize;
```
cfc4n, line 352: If only the removal of options of type `TCP_OPT_ADDR` is done here, it still cannot avoid the problem of losing the source IP.
如果这里只是对`TCP_OPT_ADDR`类型的Option做移除，那么依旧无法规避源IP丢失的问题。

src/ipvs/ip_vs_proto_tcp.c
```diff
@@ -306,46 +306,129 @@ static void tcp_in_remove_ts(struct tcphdr *tcph)
...
+    pruned = payload - slow;
+    if (pruned < reqlen) {
+        /* further trim the options, the tcp functionality relies on unprotected
+         * options may get hurt, refer to:
+         *   https://www.iana.org/assignments/tcp-parameters/tcp-parameters.xhtml
+         *   #tcp-parameters-1
+         * */
+        ptr = slow;
+        slow = fast = (unsigned char *)(tcph + 1);
+        if (tcph->syn)
```
donghaobo, line 378: SYN packet do not call `tcp_in_prune_options`

src/ipvs/ip_vs_proto_tcp.c
```diff
@@ -306,46 +306,129 @@ static void tcp_in_remove_ts(struct tcphdr *tcph)
...
+            opts_protected = (1ULL << 2) | (1ULL << 3) | (1ULL << 4)    /* MSS, WS, SACKP */
+                | (1ULL << 8) | (1ULL << 30) | (1ULL << 34);            /* TS, MPTCP, TFO */
+        else
+            opts_protected = (1ULL << 8);    /* TS, drop SACK, MPTCP DSS/REMOVE_ADDR */
```
donghaobo, line 382: TIMESTAMP was removed in SYN packet, should the packets contain TIMESTAMP option?

src/ipvs/ip_vs_proto_tcp.c
ywc689: Yes. Syn packet won't arrive here any more.  `tcp_in_prune_options` is designed to be independent of the caller, so the syn is also concerned here.

src/ipvs/ip_vs_proto_tcp.c
ywc689: It exists for similar reason as the above.

src/ipvs/ip_vs_proto_tcp.c
```diff
@@ -306,46 +306,129 @@ static void tcp_in_remove_ts(struct tcphdr *tcph)
...
+        l3hdr = rte_pktmbuf_mtod(mbuf, void *);
+        if (unlikely(mbuf_may_pull(mbuf, mbuf->pkt_len) != 0)) {
+            memset(slow, 0, pruned);
+            return EDPVS_INVPKT;
+        }
+        if (unlikely(payload - l3hdr > mbuf->pkt_len)) {
+            memset(slow, 0, pruned);
+            return EDPVS_INVPKT;
+        }
+        memmove(slow, payload, mbuf->pkt_len - (payload - l3hdr));
```
donghaobo, line 420: leave reqlen bytes for tcp_in_add_toa ?

ywc689: ` tcp_in_prune_options` tries to find `reqlen`  bytes in tcp option space, and stops immediately after achieving the goal.  Then the memory after `slow` pointer in tcp option  is invalid and should be trimmed.

ywc689: Yes, reserving reqlen bytes spares one memory shift later in toa inserting. But I think it require changes in the `tcp_in_add_toa ` function. I'd not like to do more changes, so trimming all bytes released from tcp option here.

PR #937, opened 2024-01-24
Fix backend update racing problem with dpvs-agent and healthcheck.

ywc689:


you-looks-not-tasty (approved):

tools/dpvs-agent/cmd/ipvs/get_vs.go
```diff
@@ -40,18 +40,28 @@ func NewGetVs(cp *pool.ConnPool, parentLogger hclog.Logger) *getVs {
 }
 
 func (h *getVs) Handle(params apiVs.GetVsParams) middleware.Responder {
+	shareSnapshot := settings.ShareSnapshot()
+	if params.Healthcheck != nil && !*params.Healthcheck {
+		return apiVs.NewGetVsOK().WithPayload(shareSnapshot.GetModels(h.logger))
+	}
+
+	// if params.Snapshot != nil && *params.Snapshot {
+	//	shareSnapshot.DumpTo(settings.LocalConfigFile(), h.logger)
```
liuxuting, line 49: why do you remove DumpTo? And how to generate cache file?

you-looks-not-tasty: The default value of `snapshot` is true, and the dpvs-healthcheck module has not been set `snapshot` to false. 

healthcheck module makes heavily invoked to getVs(), which will generate a large number of cache files.

We will fix this issue in the next version.

PR #938, opened 2024-01-31
use better random source

donghaobo:
use better random source


ywc689:
Could you please evaluate the performance differences between the old and the new random sources?

donghaobo:
> Could you please evaluate the performance differences between the old and the new random sources?

`generate_random_key` is only called at startup, there is no difference in packet processing.
`open` `read` and `close` take about 300us

ywc689 (approved):

PR #979, opened 2024-07-10
Multicast address bugfix and improvements

ywc689:
1.  fix hw multicast address sync problems
* Neglected the fact that multicast IPv4/IPv6 address can be mapped to one multicast hw address, and a lower dpvs port may have multiple upper ports (such as vlan).
* Multicast hw addresses could sync from kni multiple times or be deleted by mistake.
* Interferences of linked down kni devices.
 
2. Add supports for getting dpvs multicast via `dpip maddr`.


ytwang0320 (approved):
approve +1

PR #1047, opened 2025-08-26
fix: add null pointer checks in route_new_entry to prevent crashes

victory460:
To  avoid crashes when `new_route->mtu = port->mtu;`


ytwang0320 (approved):
approve +1

PR #1052, opened 2025-09-15
ip_vs_service: fix add service failed when vip like 1:1:1:1:1:1:1:1

xinchizhe:
When vip is IPv6 and every in6.s6_addr32 is same(for example 1:1:1:1:1:1:1:1), inet_addr_fold will return 0, be considered as an error(bad ip protocol)


ywc689 (approved):

ytwang0320 (approved):
approve +1

PR #1055, opened 2025-10-28
keepalived: correct inverted kill return check in MISC_CHECK

lxchx:
The existing logic incorrectly interpreted the non-zero return value
(failure) as success, leading to an inverted decision in the MISC_CHECK
process. This commit corrects the check to properly identify kill
failures.

Reference: https://github.com/acassen/keepalived/commit/2738ee60


ytwang0320 (approved):
Approve +1

